PR #5, opened 2018-05-14
Add package release script for pcdshub

ZLLentz:
## Description

```
pkg_release pmgr R1.0.2
```
Clones the repo and creates a `R1.0.2` directory in the current directory. Does reasonable error checking. Primarily for things that have not been given conda-recipe support yet.

## Motivation and Context

Nolan did this manually today, it should be scripted.

## How Has This Been Tested?

I tested it and various error cases

It hasn't


ZLLentz:
@teddyrendahl thoughts on `softlink_update` script (with a better name)?

PR #12, opened 2019-03-14
ENH: Set environment variables for QtDesigner and Happi

teddyrendahl:


ZLLentz:
@teddyrendahl is it time for tag/release following this PR?

PR #90, opened 2021-12-09
various changes to pedestals

silkenelson:
Use the new calibration commands for the Jungfrau & fix epix10k pedestals to work on FFB
## Description
Not much more details needed.

## Motivation and Context
The epix10k pedestals should be processable on the FFB as the data is available (potentially a lot) faster.
The Jungfrau pedestals should use the new scripts. This also allows to process calibration cycles in parallel and deploys a other files that currently need to be added by hand.

## How Has This Been Tested?
test from development directory on a number of runs, using both interactive& queue modes both on psana and psffb.
Not really.

### DIFF
--- a/scripts/makepeds
+++ b/scripts/makepeds
@@ -38,7 +38,7 @@ OPTIONS:
                 dark run for XTCAV
         -L|--xtcav_lasingoff
                lasing off run for XTCAV
-        -v|--valdity_start <val_run>
+        -v|--validity_start <val_run>
                validity range (set to <val_run>-end)
         -N|--nevents <#>  
                 use this number of events (default 1000). Needed when using -c as original events are counted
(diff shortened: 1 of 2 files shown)


vespos (approved):
I don't see anything wrong with these changes!

scripts/makepeds
```diff
@@ -38,7 +38,7 @@ OPTIONS:
                 dark run for XTCAV
         -L|--xtcav_lasingoff
                lasing off run for XTCAV
-        -v|--valdity_start <val_run>
+        -v|--validity_start <val_run>
```
ZryletTC, line 41: Doesn't this need to be reflected in line 178 of makepeds_psana as well?

silkenelson:
Pulling this now. Add other changes on top later.

PR #91, opened 2021-12-10
DEV: Add Beamline motors Report Script

JuliethOteroRodriguez:
**Description**
Creation of a new script that checks the IMS motors configuration in the parameter manager.

**Motivation and Context**
This script helps the engineers and scientists to know if the IMS motors' configuration status.

The script will generate a report for:
1. Motors with no configuration.
2. Motors with configurations that have values different.
from the configured.
3. Motors are not found.
4. Full report with all the motors and their status (If they have configuration assigned if that configuration matches or not with the configured values,  and if the motors are not found).

**How Has This Been Tested?**

The script was tested in the hutches XPP, CXI, MFX, MEC, XCS. The parameter manager was also another resource to check if the information generated by the script is coherent. 

**How Has This Been Documented?**

The script contains comments with the description per each function.

### DIFF
--- a/scripts/BLmotors
+++ b/scripts/BLmotors
@@ -193,12 +193,12 @@ def motorsInfo(motors, motorsName, cfg):
                     infoperRecord.append(mconf)
                     infoperRecord.append("Yes")
 
-            except:
+            except Exception:
                 infoperRecord.append(motors[motor])
                 infoperRecord.append(motorPtwo)
                 infoperRecord.append("None")
                 infoperRecord.append(" ")
-        except:
+        except Exception:
             try:
                 mconf = pm.get_config(motors[motor])
                 exConfig = pm.diff_config(motors[motor])
@@ -212,7 +212,7 @@ def motorsInfo(motors, motorsName, cfg):
                     infoperRecord.append("No found!")
                     infoperRecord.append(mconf)
                     infoperRecord.append("Yes")
-            except:
+            except Exception:
                 infoperRecord.append(motors[motor])
                 infoperRecord.append("No found!")
                 infoperRecord.append(" ")
--- a/scripts/BLmotors
+++ b/scripts/BLmotors
@@ -1,17 +1,19 @@
 #!/reg/g/pcds/pyps/conda/py36/envs/pcds-4.1.6/bin/python
 
-
 """
 Beam Line motors Report
 This script helps the engineers and scientists to know
-if the beam line motors have a configuration saved in the
-parameter manager.
+if the IMS motors configuration status.
 
 The script will generate a report for:
-1. The IMS motors with their rela PV name.
-2. The current configuration on it.
-3. If the current configuration matches or not
-with the appripite one.
+1. Motors with no configuration.
+2. Motors with configurations that have values different
+from the configured.
+3. Motors are not found.
+4. Full report with all the motors and their status
+(If they have configuration assigned, if that configuration
+matches or not with the configured values,  and if the
+motors are not found).
 """
 import os
 import argparse
--- a/scripts/BLmotors
+++ b/scripts/BLmotors
@@ -1,20 +1,17 @@
 #!/reg/g/pcds/pyps/conda/py36/envs/pcds-4.1.6/bin/python
 
-"""
-Beamline Motors Report
 
+"""
+Beamline motors Report
 This script helps the engineers and scientists to know
-if the IMS motors configuration status.
+if the beamline motors have a configuration saved in the
+parameter manager.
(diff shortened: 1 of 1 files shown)


scripts/BLmotors
```diff
@@ -0,0 +1,311 @@
...
+                if exConfig:
+                    infoperRecord.append(motors[motor])
+                    infoperRecord.append(motorPtwo)
+                    infoperRecord.append(mconf)
+                    infoperRecord.append("No")
+                else:
+                    infoperRecord.append(motors[motor])
+                    infoperRecord.append(motorPtwo)
+                    infoperRecord.append(mconf)
+                    infoperRecord.append("Yes")
```
ZLLentz, line 197: There's a bit of duplicated code here, two cases where we do the same thing on each side.
```suggestion
                infoperRecord.append(motors[motor])
                infoperRecord.append(motorPtwo)
                infoperRecord.append(mconf)
                if exConfig:
                    infoperRecord.append("No")
                else:
                    infoperRecord.append("Yes")
```

scripts/BLmotors
```diff
@@ -0,0 +1,311 @@
...
+                if exConfig:
+                    infoperRecord.append(motors[motor])
+                    infoperRecord.append("No found!")
+                    infoperRecord.append(mconf)
+                    infoperRecord.append("No")
+                else:
+                    infoperRecord.append(motors[motor])
+                    infoperRecord.append("No found!")
+                    infoperRecord.append(mconf)
+                    infoperRecord.append("Yes")
```
ZLLentz, line 217: Same as above, some duplication here:
```suggestion
                infoperRecord.append(motors[motor])
                infoperRecord.append("No found!")
                infoperRecord.append(mconf)
                if exConfig:
                    infoperRecord.append("No")
                else:
                    infoperRecord.append("Yes")
```

scripts/BLmotors
```diff
@@ -0,0 +1,311 @@
...
+    This function displays in a table the information
+    of all the ims motors by selection of the user.
+    """
+    noConf = [info for info in totalinfo if info[0] == info[1]
+              and info[2] == "None" or info[0] != info[1]
+              and info[3] != "None" and info[1] != "No found!"]
+    noMatchConf = [info for info in totalinfo if info[3] == "No"]
+    noFound = [info for info in totalinfo if info[1] == "Not found!"]
+    aliasConf = [info for info in totalinfo if info[0] != info[1]
+                 and info[1] != "No found!" and info[2] != "None"]
```
ZLLentz, line 239: I suspect the above logic was difficult to debug. This is complicated enough to switch to using dictionaries or dataclasses so you can compare named fields instead of array indices, which would make this much easier to understand.

scripts/BLmotors
```diff
@@ -0,0 +1,311 @@
...
+            tableNoconf.field_names = ["Record Name", "Pv Name",
+                                       "Configuration",
+                                       "Configuration Matches"]
+            tableNoconf.add_rows(sortedListNoconf)
+            print(tableNoconf)
+        else:
+            print('The IMS motors have a configuration or/and',
+                  ' the IMS motors were not found!')
+
+    elif argMatch:
```
ZLLentz, line 255: The `elif` means that only one of the options can be used at a time, right? I suggest either changing these all to `if` so the user can see multiple reports at once or writing in the help text that only one can be used at a time.

ZLLentz (commented):
I wrote a lot of review here, the most important thing is the default value for PYPS_ROOT, else the script won't run for all users. The rest are mostly suggestions for readability since a lot of this is confusing.

scripts/BLmotors
```diff
@@ -0,0 +1,311 @@
...
+    actual configuration. Then it will create a list
+    per each record and finally it will be addes to
+    another list.
+    """
+    pm = pmgrAPI.pmgrAPI("ims_motor", cfg)
+    totalinfo = []
+    for motor in range(len(motors)):
+        infoperRecord = []
+        try:
+            motorPtwo = ophyd.signal.EpicsSignal(motorsName[motor]).get()
```
ZLLentz, line 184: One more thing to note for the future: this line is likely the speed bottleneck for the script. To make it run faster (if that is needed), the EPICS mechanisms would need to be parallelized.

ZryletTC (approved):
I'll approve since it's working and it's the end of the year. I'll add all unresolved review comments to an issue which can be addressed in another PR.

ZLLentz:
I think it's totally fine to merge if the PYPS root default was added, I just wanted to document all the little things.

PR #94, opened 2022-01-13
Changes to camViewer

### LINKED ISSUE #76: ENH: Future improvements for camViewer
1. Reference imgr.py from within engineering_tools
2. Add full worded options (--enable/--disable) in addition to the the single letter option (-e/-d)
3. Troubleshoot problem with accessing cameras outside of hutch (lfe/fee)

spenc333:
## Description

Changes to camviewer script involved making fully worded enable and disable options as well fixing the -H option that now allows for a camera to be accessed from any hutch. Adding the imgr script to eng_tools instead of its original directory allows for control of an ioc given it's name and hutch . Also needed this script to be called in camviewer

## Motivation and Context

Fixes #76

## How Has This Been Tested?

Tests involved sourcing the script from my epics-dev directory using various camera names and hutches as well as the other options. Since imgr is in development and needed this command within camviewer, it also had to be sourced from epics-dev directory within the camviewer script.

Both scripts have updated usage functions


scripts/camViewer
```diff
@@ -258,15 +274,44 @@ fi
 
 if [ $ENABLE -gt 0 ]; then 
     echo "Enabling ${CAMNAME}"
-    IOC=$(iocfpv "${CAMPVFULL}")
-    $imgr "$IOC" --hutch "$hutch" --enable   
+    for ioc in `iocfpv "$CAMPVFULL:IMAGE1"`; do
```
ZLLentz, line 277: Can you briefly explain what these loops are fixing, compared to the previous implementation? I'm not familiar with this `iocfpv` command or the implications of doing this multiple times.

scripts/camViewer
```diff
@@ -258,15 +274,44 @@ fi
...
+
+    imgr "$ioc" --hutch "$iochutch" --enable   
     exit
 fi
 
 if [ $DISABLE -gt 0 ]; then
     echo "Disabling ${CAMNAME}"
-    IOC=$(iocfpv "${CAMPVFULL}")
-    $imgr "$IOC" --hutch "$hutch" --disable    
+    for ioc in `iocfpv "$CAMPVFULL:IMAGE1"`; do
```
ZLLentz, line 299: same comments/questions about the second loop as for the first loop, but also perhaps there was a way to combine these two into one section to avoid code duplication?

ZLLentz (commented):
Added a bunch of comments/questions/suggestions

scripts/imgr
```diff
@@ -0,0 +1,63 @@
...
+   exit 0
+fi
+
+ulimit -c unlimited 
+
+source /reg/g/pcds/setup/pathmunge.sh
+ldpathmunge /usr/local/lib
+
+DIRTMP=`dirname  "${BASH_SOURCE[0]}"`
+DIR="$( cd $DIRTMP && pwd -P )"
```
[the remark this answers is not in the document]
ZryletTC: No I don't think it does since the command is run in a subshell. It's still probably cleaner to use `DIR=$(readlink -f $(dirname  "${BASH_SOURCE[0]}")) though`

scripts/camViewer
spenc333: Since a lot of the fee cameras have multiple iocs associated with a single PV this loops over all the iocs for a given PV and picks the relevant one.

scripts/camViewer
spenc333: I will attempt to create a function that can be called in both.

scripts/imgr
```diff
@@ -0,0 +1,63 @@
...
+
+if [[ $1 == *'-'* ]]; then
+    if [[ $# -eq 4 ]]; then
+        /reg/g/pcds/pyps/apps/iocmanager/latest/imgr "$1" --hutch $3 "$4"  > /dev/null &
+    elif [[ $5 == *':'* && $# -eq 5 ]]; then
+        /reg/g/pcds/pyps/apps/iocmanager/latest/imgr "$1" --hutch $3 "$4" "$5" > /dev/null &
+    elif [[ $3 == *'list'* && $# -eq 6 ]]; then
+        /reg/g/pcds/pyps/apps/iocmanager/latest/imgr --hutch $2 --list --host "$5" "$6"
+    elif [[ $# -eq 5 && $5 != *':'* || $3 != *'list'* ]]; then
+        /reg/g/pcds/pyps/apps/iocmanager/latest/imgr "$1" --hutch $3 "$4" "$5" > /dev/null &
```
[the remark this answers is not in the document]
spenc333: This is assuming the user follows the syntax in the usage function. If they don't you get some error messages from this file as well in some cases from the original script.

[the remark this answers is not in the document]
ZryletTC: Would it be possible to get rid of all this logic and just pass the given arguments to imgr and let it deal with them?

scripts/camViewer
```diff
@@ -258,15 +274,44 @@ fi
...
     echo "Enabling ${CAMNAME}"
-    IOC=$(iocfpv "${CAMPVFULL}")
-    $imgr "$IOC" --hutch "$hutch" --enable   
+    for ioc in `iocfpv "$CAMPVFULL:IMAGE1"`; do
+        
+        INFO=$(grep_ioc "$ioc" all | grep "id:'$ioc'")
+        if [ -z "$INFO" ]; then
+             
+            >&2
+        else
```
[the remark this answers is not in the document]
spenc333: I wanted something that allowed me to loop over the iocs without stopping the 'for' loop if a given ioc produced no information about it in iocmanager.

scripts/camViewer
```diff
@@ -152,29 +154,17 @@ if [[ ($1 == "--help") || ($1 == "-h") ]]; then
...
 WAIT=0
 RATE=-1
 REBOOT=0
 ENABLE=0
 DISABLE=0
+MAINSCREEN=0
 
 
-while getopts "c:w:u:H:lmred" OPTION
+while getopts "c:w:u:H:lmred":"-enab":"-disab" OPTION
```
ZryletTC, line 167: Does that work for getopts? If not, it might be easier to get rid of getopts and loop through the arguments the old-fashioned way

scripts/camViewer
```diff
@@ -212,9 +202,34 @@ do
         d)
             DISABLE=1
             ;;
+        -disable)
+            DISABLE=1
+            ;;
+        -enable)
+            ENABLE=1
+            ;;
```
ZryletTC, line 210: These can be combined into the corresponding cases above
eg. `-e|--enable)`

scripts/camViewer
```diff
@@ -258,15 +274,44 @@ fi
...
+
+        INFO=$(grep_ioc "$ioc" all | grep "id:'$ioc'")
+        if [ -z "$INFO" ]; then
+
+            >&2 
+        else
+
+            IOC=$(echo "$INFO" | sed -n "s/^.*id: '\(\S*\)'.*$/\1/p")
+            IOCHOST=$(echo "$INFO" | sed -n "s/^.*host: '\(\S*\)'.*$/\1/p")
+            iochutch=$(echo $INFO | grep "/reg/g/pcds/pyps/config/*" | cut -d'/' -f7)
```
ZryletTC, line 309: Is the grep in this line necessary?

scripts/camViewer
silkenelson: If you mean to say we still have traces of earlier implementations for a given camera, I'd suggest cleaning this up. If that is not what you meant, can you expand on why we have multiple IOCs associated with a given <BASE>:IMAGE1 PV?

scripts/camViewer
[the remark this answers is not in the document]
ZryletTC: So can you just invert the condition and get rid of the empty side?

scripts/camViewer
spenc333: Yes, I meant to say that we have traces of earlier implementations.

scripts/camViewer
spenc333: I'm not sure how else I would search for that path within $INFO.

ZryletTC: Doesn't every line in INFO have that string in it?

spenc333: There's only one instance of it. What do you mean?

scripts/imgr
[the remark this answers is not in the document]
spenc333: It doesn't when I test with other options

scripts/imgr
```diff
@@ -0,0 +1,63 @@
...
+ldpathmunge /usr/local/lib
+
+DIRTMP=`dirname  "${BASH_SOURCE[0]}"`
+DIR="$( cd $DIRTMP && pwd -P )"
+pathmunge $DIR
+
+
+if [[ $1 == *'-'* ]]; then
+    if [[ $# -eq 4 ]]; then
+        /reg/g/pcds/pyps/apps/iocmanager/latest/imgr "$1" --hutch $3 "$4"  > /dev/null &
```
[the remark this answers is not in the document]
spenc333: There was no special reason  and it doesn't matter if I use it or not, I will get rid of it.

scripts/imgr
```diff
@@ -0,0 +1,63 @@
...
+        /reg/g/pcds/pyps/apps/iocmanager/latest/imgr "$1" --hutch $3 "$4" "$5" > /dev/null &
+    elif [[ $3 == *'list'* && $# -eq 6 ]]; then
+        /reg/g/pcds/pyps/apps/iocmanager/latest/imgr --hutch $2 --list --host "$5" "$6"
+    elif [[ $# -eq 5 && $5 != *':'* || $3 != *'list'* ]]; then
+        /reg/g/pcds/pyps/apps/iocmanager/latest/imgr "$1" --hutch $3 "$4" "$5" > /dev/null &
+    else
+        echo 'Conditions not met'
+    fi
+else
+    echo 'Ioc does not exist' >&2
```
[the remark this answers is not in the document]
spenc333: Why not?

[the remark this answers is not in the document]
ZryletTC: Because it's checking for a hyphen, not actually searching for the ioc

scripts/imgr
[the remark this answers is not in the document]
ZryletTC: What part doesn't work? From a few quick tests, the following seems to work ok:
```
#!/bin/bash
for (( i=1; i<=$#; i++)); do
    if [[ "${!i}" == "--hutch" ]]; then
        j=$((i+1))
        break
    fi
done

HUTCH=${!j:-`get_hutch_name`}

/reg/g/pcds/pyps/config/${HUTCH}/iocmanager/imgr $@
```

scripts/camViewer
```diff
@@ -127,6 +127,18 @@ iocfpv() {
...
 
+iocinfo() {
+     for ioc in `iocfpv "$CAMPVFULL:IMAGE1"`; do
+
+        INFO=$(grep_ioc "$ioc" all | grep "id:'$ioc'")
+        if [ -z "$INFO" ]; then
+            :            
+        else
+            iochutch=$(echo $INFO | cut -d'/' -f7)
+        fi
```
ZLLentz, line 138: More compact way to handle these sorts of cases. `-n` is effectively the opposite of `-z`
```suggestion
        if [ -n "$INFO" ]; then
            iochutch=$(echo $INFO | cut -d'/' -f7)
        fi
```

scripts/camViewer
ZryletTC: Is actually resolved?

spenc333: Yes.

scripts/camViewer
```diff
@@ -206,15 +208,31 @@ do
...
             DISABLE=1
             ;;
 	esac
 done
 
+
+if [ "$hutch" == ""  ]; then
+    hutch=`get_info --gethutch`
+    if [ "$hutch" == unknown_hutch ]; then
+        echo 'unknown hutch, cannot use camViewer script'
```
ZryletTC, line 224:
```suggestion
        echo 'unknown hutch, specify using -H option'
```

scripts/camViewer
```diff
@@ -206,15 +208,31 @@ do
...
+        exit 1
+    fi
+fi
+
+
+
+EXE=/reg/g/pcds/pyps/config/$hutch/camviewer/run_viewer.csh
+PVLIST=/reg/g/pcds/pyps/config/$hutch/camviewer.cfg
```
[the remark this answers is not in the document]
ZryletTC: It also doesn't seem to affect the list when using the `-l` option. I wouldn't consider the -H option fixed yet.

scripts/camViewer
```diff
@@ -152,29 +165,18 @@ if [[ ($1 == "--help") || ($1 == "-h") ]]; then
...
 WAIT=0
 RATE=-1
 REBOOT=0
 ENABLE=0
 DISABLE=0
+MAINSCREEN=0
 
 
-while getopts "c:w:u:H:lmred" OPTION
+while getopts " -o c:w:u:H:lmre:(enable)d:(disable)" OPTION
```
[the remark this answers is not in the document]
spenc333: https://docs.oracle.com/cd/E19253-01/816-5165/getopts-1/index.html

scripts/imgr
```diff
@@ -0,0 +1,45 @@
...
+ulimit -c unlimited 
+
+source /reg/g/pcds/setup/pathmunge.sh
+ldpathmunge /usr/local/lib
+
+DIR=$(readlink -f $(dirname "${BASH_SOURCE[0]}"))
+PATH=$PATH:$DIR
+pathmunge $DIR
+
+/reg/g/pcds/pyps/apps/iocmanager/latest/imgr $@
```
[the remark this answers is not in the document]
spenc333: Why is that?

ZLLentz:
One more suggestion for any time you're doing a bash script: make sure you run the result through shellcheck, it can often help catch issues.
There's a `shellcheck` utility in pcds_conda and also a website https://www.shellcheck.net/

scripts/imgr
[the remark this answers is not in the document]
ZryletTC: Otherwise, what's the point of having different versions for each hutch? Maybe a new release will break things in one hutch so they will want to stick with an older version.

scripts/imgr
```diff
@@ -0,0 +1,45 @@
...
+fi
+
+ulimit -c unlimited 
+
+source /reg/g/pcds/setup/pathmunge.sh
+ldpathmunge /usr/local/lib
+
+DIR=$(readlink -f $(dirname "${BASH_SOURCE[0]}"))
+PATH=$PATH:$DIR
+pathmunge $DIR
```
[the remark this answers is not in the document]
spenc333: I guess you're right. I believe that here PATH and pathmunge do the same thing

scripts/imgr
```diff
@@ -0,0 +1,45 @@
...
+   echo 'need arguments' >&2
+   usage
+   exit 0
+fi
+
+
+if [[ ($1 == "--help") || ($1 == "-h") ]]; then
+   usage
+   exit 0
+fi
```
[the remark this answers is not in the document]
spenc333: Sure, I can make that change

scripts/camViewer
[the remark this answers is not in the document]
spenc333: How are you calling the -l?  "camViewer -H $hutch -l"

silkenelson:
It feels wrong to me to add the imgr script to engineering tools as long as it also lives in the iocmanager package - with exactly the same name.

spenc333:
> It feels wrong to me to add the imgr script to engineering tools as long as it also lives in the iocmanager package - with exactly the same name.

I agree.

ZryletTC:
@silkenelson This is pretty much how iocmanager is done as well, do you have a better alternative? At the moment, there is no default way to use imgr than by specifying the full path to the executable. Adding it to engineering_tools adds it to everyone's PATH without extending the PATH variable.
I brought up the question of how to handle this with you and Mike over Slack back in June. He agreed that a wrapper script in engineering_tools was probably best but I never heard your opinion. If there's a better way, I'm all ears.

silkenelson:
If I was part of that discussion, I have forgotten - I am unable to retain slack discussions for much longer than a week. Yes indeed: the answer that removing this script from iocmanager was part of this request (or at least where this was going) answers my question.

ZryletTC:
Once #101 is merged, the imgr addition should be removed from this script, then I'll rereview this PR and we can hopefully merge.

ZLLentz:
I'm excited to use this one, I semi-frequently find myself wishing I had the hutch selector here

ZryletTC (changes requested):
Check the code, the style fixes have broken some functionality.

scripts/camViewer
```diff
@@ -176,38 +177,52 @@ DISABLE=0
 MAINSCREEN=0
 
 
-while getopts " -o c:w:u:H:lmre:(enable)d:(disable)" OPTION
+OPTIONS=$(getopt -o lmredc:w:u:H: --long list,main,reboot,enable,disable,cam:,wait:rate:hutch: -- "$@")
```
ZryletTC, line 180:
```suggestion
OPTIONS=$(getopt -o lmredc:w:u:H: --long list,main,reboot,enable,disable,cam:,wait:,rate:,hutch: -- "$@")
```

scripts/camViewer
```diff
@@ -148,26 +148,27 @@ usage: $0 options
...
--m:            bring up the edm screen
--r:            reboot the IOC
--l:            print list of cameras
--w #:         (wait for # hours to ask to renew, default 2 and 12 for GIGEs)
--u #:          update rate limit (default 5)
--H hutch:      use a specific hutches camviewer config file
--e|--enable:   enable camera ioc
--d|--disable:  disable camera ioc
+-c|--cam    :   camera name as in camera list or gige #
+-m|--main   :   bring up the edm screen
```
ZryletTC, line 152:
```suggestion
-m|--main   :   bring up the edm screen
```
@silkenelson is that what the -m stands for?

scripts/camViewer
```diff
@@ -148,26 +148,27 @@ usage: $0 options
...
+-r|--reboot :   reboot the IOC
+-l|--list   :   print list of cameras
+-w|--wait # :   (wait for # hours to ask to renew, default 2 and 12 for GIGEs)
+-u|--rate # :   update rate limit (default 5)
```
ZryletTC, line 156:
```suggestion
-u|--rate # :   update rate limit in Hz (default 5)
```

scripts/camViewer
```diff
@@ -148,26 +148,27 @@ usage: $0 options
...
--w #:         (wait for # hours to ask to renew, default 2 and 12 for GIGEs)
--u #:          update rate limit (default 5)
--H hutch:      use a specific hutches camviewer config file
--e|--enable:   enable camera ioc
--d|--disable:  disable camera ioc
+-c|--cam    :   camera name as in camera list or gige #
+-m|--main   :   bring up the edm screen
+-r|--reboot :   reboot the IOC
+-l|--list   :   print list of cameras
+-w|--wait # :   (wait for # hours to ask to renew, default 2 and 12 for GIGEs)
```
ZryletTC, line 155:
```suggestion
-w|--wait # :   wait for # hours to ask to renew, default 2 (12 for GIGEs)
```

scripts/camViewer
```diff
@@ -148,26 +148,27 @@ usage: $0 options
...
+-u|--rate # :   update rate limit (default 5)
+-H|--hutch  :   use a specific hutches camviewer config file
```
ZryletTC, line 157:
```suggestion
-c|--cam    :   camera name as in camera list or gige #
-m|--main   :   bring up the edm screen
-r|--reboot :   reboot the IOC
-l|--list   :   print list of cameras
-w|--wait   :   (wait for # hours to ask to renew, default 2 and 12 for GIGEs)
-u|--rate   :   update rate limit (default 5)
-H|--hutch  :   use a specific hutches camviewer config file
```
for consistency. But probably clearer to add arguments after all options that require values instead.

scripts/camViewer
```diff
@@ -176,38 +177,52 @@ DISABLE=0
...
+        -e|--enable)
             ENABLE=1
+            shift
             ;;
-        d|disable)
+        -d|--disable)
             DISABLE=1
+            shift
+            ;;
+        --) shift
```
ZryletTC, line 221: Can you please add a warning for unrecognized arguments?

ZryletTC (approved):
Cool, looks good, all my tests have worked :) (after adding the commas)

scripts/camViewer
spenc333: From what I've seen it is

ZryletTC: I guess it is... 
Kinda feel like most users would call the camviewer the "main" screen and that the option should be --edm or something though

scripts/camViewer
spenc333: That's what line 182 does.

ZryletTC: Ok, is it able to handle options without a `-` in front of it?
For example, `./camViewer blah -H xcs` proceeds without acknowledging the `blah`

scripts/camViewer
```diff
@@ -135,98 +148,124 @@ usage: $0 options
...
+        --) shift 
+            break 
             ;;
-	esac
+        
+       
+    esac
 done
 
+
```
ZryletTC, line 229:
```suggestion
if [ $# -ge 1 ]; then
    echo "Positional arguments unsupported: $@" >&2
    usage
    exit 1
fi

```

PR #101, opened 2022-02-07
Add imgr and merge iocmanager into one bash file

ZryletTC:
## Description

- Add imgr
- Change iocmanager to bypass the IocManager bash script in the IocManager package

## Motivation and Context

The bash scripts to access imgr and iocmanager should exist in one place only and engineering_tools is a convenient place because it houses many other commonly used scripts and is already on most peoples's PATHs. This is the first step in moving the scripts from the iocmanager package to engineering_tools. After all hutches have adopted a new release of engineering_tools containing these scripts, I will remove the no-longer-useful bash scripts from the iocmanager package.

## How Has This Been Tested?
Interactively.
Shellcheck passes with only SC1091 messages.

Commented each section in both scripts.

### DIFF
--- a/scripts/imgr
+++ b/scripts/imgr
@@ -1,5 +1,35 @@
 #!/bin/bash
 
+# Define usage here for faster return
+usage()
+{
+cat << EOF
+usage: imgr <IOCNAME> [--hutch HUTCH] <OPTION>
+
+Control status of all IOCs running in a particular hutch from the command line.
+Current hutch is used if not provided.
+
+List of options:
+imgr IOCNAME [--hutch HUTCH] --reboot soft
+imgr IOCNAME [--hutch HUTCH] --reboot hard
+imgr IOCNAME [--hutch HUTCH] --enable
+imgr IOCNAME [--hutch HUTCH] --disable
+imgr IOCNAME [--hutch HUTCH] --upgrade RELEASE_DIR
+imgr IOCNAME [--hutch HUTCH] --move HOST
+imgr IOCNAME [--hutch HUTCH] --move HOST:PORT
+imgr [--hutch HUTCH] --list [--host HOST] [--enabled_only|--disabled_only]
+ 
+EOF
+}
+if [ $# -lt 1 ]; then
+    echo 'Missing required arguments' >&2
+    usage
+    exit 1
+elif [[ ($1 == "--help") || ($1 == "-h") ]]; then
+    usage
+    exit 0
+fi
+
 # Setup environment
 if [ -x /etc/pathinit ]; then
     source /etc/pathinit
--- a/scripts/imgr
+++ b/scripts/imgr
@@ -74,5 +74,5 @@ if [[ "$HUTCH" == "unknown_hutch" ]]; then
     echo "Unknown hutch, please use --hutch argument" >&2
     exit 1
 else
-    /reg/g/pcds/pyps/config/${HUTCH}/iocmanager/imgr.py "$@" --hutch $HUTCH
+    /reg/g/pcds/pyps/config/"${HUTCH,,}"/iocmanager/imgr.py "$@" --hutch "${HUTCH,,}"
 fi
--- a/scripts/imgr
+++ b/scripts/imgr
@@ -1,0 +1,44 @@
+#!/bin/sh
+
+# Setup environment
+if [ -x /etc/pathinit ]; then
+    source /etc/pathinit
+else
+    export PSPKG_ROOT=/cds/group/pcds/pkg_mgr
+    export PYPS_ROOT=/cds/group/pcds/pyps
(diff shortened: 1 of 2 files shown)


scripts/imgr
```diff
@@ -0,0 +1,78 @@
...
+        break
+    fi
+done
+
+# Hutch choice priority order:
+# 1. --hutch arg
+# 2. HUTCH environment variable
+# 3. get_hutch_name
+HUTCH=${ARGHUTCH:-$HUTCH}
+HUTCH=${HUTCH:-$(get_hutch_name)}
```
ZLLentz, line 70: Does bash handle this smartly, e.g. skipping execution of `get_hutch_name` if `$HUTCH` is set? (I know that it doesn't use the output, but does it also not run the code?)

scripts/imgr
```diff
@@ -0,0 +1,78 @@
...
+    echo "IocManager 2.0.0 and higher does not run on RHEL5!" >&2
+    exit 1
+fi
+export QT_XKB_CONFIG_ROOT=/usr/share/X11/xkb
+export PSPKG_RELEASE=controls-0.1.9
+source $PSPKG_ROOT/etc/set_env.sh
+DIR=$(readlink -f "$(dirname "${BASH_SOURCE[0]}")")
+PATH=$PATH:$DIR
+
+# Search for hutch in args
```
ZLLentz, line 56: Is it correct to set up the environment before processing args? Minor point I guess, setting up the environment is probably pretty quick here and the script is pretty small. In a larger script this might be an issue.

scripts/imgr
```diff
@@ -0,0 +1,78 @@
...
+
+# Hutch choice priority order:
+# 1. --hutch arg
+# 2. HUTCH environment variable
+# 3. get_hutch_name
+HUTCH=${ARGHUTCH:-$HUTCH}
+HUTCH=${HUTCH:-$(get_hutch_name)}
+
+# Run if possible
+if [[ "$HUTCH" == "unknown_hutch" ]]; then
```
ZLLentz, line 73: Stuff like this makes me nervous about knock-on affects of upstream changes to `get_hutch_name`'s output. This also does not catch e.g. `--hutch xvs` typos, but it probably doesn't need to.

scripts/imgr
```diff
@@ -0,0 +1,78 @@
...
+    echo "Unknown hutch, please use --hutch argument" >&2
+    exit 1
+else
+    /reg/g/pcds/pyps/config/${HUTCH}/iocmanager/imgr.py "$@" --hutch $HUTCH
```
ZLLentz, line 77: is `$HUTCH` case-sensitive? e.g., does `XCS` vs `xcs` matter?

scripts/iocmanager
```diff
@@ -1,39 +1,58 @@
...
-    HUTCH=$1
+    /reg/g/pcds/pyps/config/"${HUTCH}"/iocmanager/IocManager.py --hutch "$HUTCH" > /dev/null &
 fi
 
-/reg/g/pcds/pyps/config/${HUTCH}/iocmanager/IocManager --hutch $HUTCH > /dev/null &
-
-if [ `whoami` == "xppopr" ]; then
-    window_id=`xdotool search --sync --onlyvisible --name 'IocManager'` 
+# Do some window cleanup for xpp
+if [ "$(whoami)" == "xppopr" ]; then
```
ZLLentz, line 54: This gave me a laugh. Maybe eventually we should have some uniform way of doing this across the various scripts and hutches.

ZLLentz (approved):
Added some minor questions/comments that apply to both files. Should be good as-is though.

scripts/imgr
ZryletTC: Couldn't find any documentation to answer that question but a quick test seems to indicate it is smart:
```
cxi-daq:current> time TEST=${TEST2:-$(get_hutch_name)}
real	0m0.058s
cxi-daq:current> TEST2=no
cxi-daq:current> time TEST=${TEST2:-$(get_hutch_name)}
real	0m0.000s
```

scripts/imgr
ZryletTC: Yeah, a quick function to check valid hutch names would be nice. Good quick issue for intern :)

scripts/imgr
ZryletTC: Yes, good point. Should I add the `${HUTCH,,}` though? It won't work on rhel5 hosts as they have an old very version of bash installed.

scripts/iocmanager
ZryletTC: Created #107

ZLLentz (approved):
LGTM

scripts/imgr
ZLLentz: Nice, that satisfies my curiosity!

PR #103, opened 2022-02-11
EpicsArchChecker (LCLS2) fixes

silkenelson:
epicsArchChecker

-detault to show only warnings, only status is optional.
-deal with the addition of ca/pva in the LCLS2 epicsArch files to properly find duplicates

## Motivation and Context
for LCLS-2, the status check currently does not do what is needed (most likely). It also makes the script run longer as it does not just read files, but actually connects to PVs.
adding the ca/pva designation for the LCLS2 epicsArch files broke the comparison
We would like to roll this out to TMO...

## How Has This Been Tested?
by running the script on the TMO epicsArch file.


ZryletTC (approved):
LGTM 👍 👍

PR #100, opened 2022-02-04
mec uses jet, cxi_0.cnf as cxi cnf file for path, only tt plugin

silkenelson:
What the title said - make ami_offline work in many more cases
## Description
MEC also uses jet
use cxi_0.cnf to figure out pathss
only time tool plugin
use the new FFB as patch

## Motivation and Context
the old version just did not work right.

## How Has This Been Tested?
tried to open a few different experiments on psana

## Where Has This Been Documented?
nowhere.

PR #104, opened 2022-02-11
MNT: Cleanup of a few files

ZryletTC:
## Description and Context

A little bit a cleanup of epicsArchChecker following #103. There is one big commit that is just capitalization changes that unfortunately makes this file a bit harder to review, sorry.

One line removed from ami_offline_psana from my [comment](https://github.com/pcdshub/engineering_tools/pull/100#discussion_r799206263) in #100.

I also could have sworn we decided to use `latest-released` as opposed to `latest` for engineering_tools. Did we change our minds at some point? This PR changes it but we can take that out if there was some reason.

Lastly, this fixes the `-m` option for restartdaq, which allows you to specify a different machine to run the daq on. This option has been around forever but I'm not sure that it ever worked. This fix makes it work but if it's not desired, we should just remove the option entirely.

## How Has This Been Tested?

Interactively in CXI

These are all fixes/cleanup commits. No features/API changed.
(except maybe the bashrc, but I don't really thinks that needs to be documented)

### DIFF
--- a/rc/bashrc
+++ b/rc/bashrc
@@ -23,7 +23,7 @@ fi
 
 source /reg/g/pcds/setup/pcds_shortcuts.sh
 
-export PATH=$PATH:/reg/g/pcds/engineering_tools/latest/scripts/
+export PATH=$PATH:/reg/g/pcds/engineering_tools/latest-released/scripts/
 
 ########################################
 ### Specific PCDS environment logins ###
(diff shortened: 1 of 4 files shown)


ZryletTC:
If the restartdaq part is accepted, restartdaq (and stopdaq) should eventually handle remote sessions a bit better. Restartdaq currently runs in a foreground ssh session that you have to Ctrl-C from to exit and stopdaq has no remote handling at the moment.

silkenelson:
I've merged this as I'd like to make a tag/release so that I can move TMO to have a working epicsArchChecker.

PR #105, opened 2022-02-14
ENH: check that the path exists when registering a module/bin

### LINKED ISSUE #62: pydev_register path verification
pydev_register should verify that the provided path is valid

_Originally posted by @ZryletTC in https://github.com/slaclab/engineering_tools/pull/60#discussion_r575102397_

nrwslac:
## Description
Added a check in the if statement that verifies the module to be registered exists.

## Motivation and Context
pydev_register will currently happily allow you to register modules that don't exist.
Closes #62 

## How Has This Been Tested?
This was tested by running the script with a module that exist and does not exist and checking to see what shows up in ~/pydev

### DIFF
--- a/scripts/pydev_register
+++ b/scripts/pydev_register
@@ -18,9 +18,9 @@ if [ -z "${1}" ]; then
   usage
   exit
 elif [ ! -e "${1}" ]; then
-  echo "Path Does Not Exist"
+  echo "Path Does Not Exist" >&2
   usage
-  exit
+  exit 1
 fi
 
 PYDEV_DIR=~/pydev
--- a/scripts/pydev_register
+++ b/scripts/pydev_register
@@ -17,7 +17,7 @@ fi
 if [ -z "${1}" ]; then
   usage
   exit
-elif [ ! -f "${1}" ]; then
+elif [ ! -e "${1}" ]; then
   echo "Path Does Not Exist"
   usage
   exit
--- a/scripts/pydev_register
+++ b/scripts/pydev_register
@@ -17,7 +17,12 @@ fi
 if [ -z "${1}" ]; then
   usage
   exit
+elif [ ! -f "${1}" ]; then
+  echo "Path Does Not Exist"
+  usage
+  exit
 fi
+
 PYDEV_DIR=~/pydev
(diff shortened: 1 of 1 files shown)


scripts/pydev_register
```diff
@@ -17,7 +17,12 @@ fi
 if [ -z "${1}" ]; then
   usage
   exit
+elif [ ! -f "${1}" ]; then
```
ZLLentz, line 20: `-f` is specifically for files, did you mean `-d` for a directory check or a `-e` for a generic existence check? I think this is typically called on directories

nrwslac: We pass the module *.py or bin file to be registered I thought. I was thinking we want to make sure these files exist?

nrwslac: the path to these files. I'm unsure, do we check the dir exist or the file using the passed path.

ZLLentz: normally I'd do something like this (with my gh clone area in my home):
```
cd ~/github/pcdsdevices
pydev_register pcdsdevices
```

ZLLentz: "Register the path `pcdsdevices` as a `module` for my dev env"

nrwslac: ah, so dont just register individual modules

ZLLentz: I think you could add a standalone .py file module, I haven't done that myself but it should work in those rare cases where you do only just have a one-file module

nrwslac: okay, looks like we want to go with `-e` here to catch individual modules and dirs.

scripts/pydev_register
```diff
@@ -17,7 +17,12 @@ fi
 if [ -z "${1}" ]; then
   usage
   exit
+elif [ ! -e "${1}" ]; then
+  echo "Path Does Not Exist"
+  usage
+  exit
```
ZryletTC, line 23:
```suggestion
  echo "Path Does Not Exist" > &2
  usage
  exit 1
```

nrwslac: updated.

ZLLentz (approved):

PR #110, opened 2022-03-02
Add simple script for checking archived PV status.

slactjohnson:
## Description

Adds a simple script for checking the status of archived PVs.

## Motivation and Context
Helpful for checking PVs in the archiver. 

## How Has This Been Tested?
Tested on some laser PVs, as well as the gas detector (see below).

A usage string.

image


scripts/check_archived_pv
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
+usage()
+{
+cat << EOF
+usage: $0
```
ZryletTC, line 5:
```suggestion
usage: $0 <PV>
```

ZryletTC: Could you please also add this to the engineering_tools README.md?

scripts/check_archived_pv
```diff
@@ -0,0 +1,19 @@
...
+
+Check on the status of a PV in the archiver, e.g. $0 GDET:FEE1:241:ENRC
+EOF
+}
+
+if [[ ($1 == "--help") || ($1 == "-h") ]]; then
+	usage
+	exit 0
```
ZryletTC, line 13: Doesn't really matter but just changing this for consistency's sake
```suggestion
    usage
    exit 0
```

scripts/check_archived_pv
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
+usage()
+{
+cat << EOF
+usage: $0
+
+Check on the status of a PV in the archiver, e.g. $0 GDET:FEE1:241:ENRC
+EOF
+}
+
```
ZryletTC, line 10:
```suggestion

if [[ -z $1 ]]; then
    usage
    exit 1
fi
    
```

scripts/check_archived_pv
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
```
ZryletTC, line 1: Are you pretty hard set on the name? Not sure about others but I think I'd prefer something like `archiveStatus`

ZryletTC (approved):
Nice addition 👍

scripts/check_archived_pv
slactjohnson: I don't really care about the name, as long as it makes sense. `archiveStatus` is just as intuitive to me. 
There's an interesting mix of naming conventions used in this repo. Looks like I picked the wrong one. Is camel case the new standard in this repo?

scripts/check_archived_pv
```diff
@@ -0,0 +1,24 @@
...
+fi
+    
+if [[ ($1 == "--help") || ($1 == "-h") ]]; then
+    usage
+    exit 0
+fi
+
+result=$(curl -s "http://pscaa01.slac.stanford.edu:17665/mgmt/bpl/getPVStatus?pv=$1")
+echo
+echo $result | cut -d'{' -f2 | cut -d'}' -f1 | tr ',' '\n'
```
ZLLentz, line 23: Not suggesting a particular change, but you should know about the `jq` utility for processing/formatting json in the terminal:
```
zlentz@mfx-control:~$ curl -s "http://pscaa01.slac.stanford.edu:17665/mgmt/bpl/getPVStatus?pv=GDET:FEE1:241:ENRC" | jq
[
  {
    "lastRotateLogs": "Never",
    "appliance": "pscaa02",
    "pvName": "GDET:FEE1:241:ENRC",
    "pvNameOnly": "GDET:FEE1:241:ENRC",
    "connectionState": "true",
    "lastEvent": "Mar/02/2022 15:14:45 -08:00",
    "samplingPeriod": "1.0",
    "isMonitored": "false",
    "connectionLastRestablished": "Feb/09/2022 13:26:40 -08:00",
    "connectionFirstEstablished": "Sep/27/2021 09:59:53 -07:00",
    "connectionLossRegainCount": "11",
    "status": "Being archived"
  }
]
```

ZLLentz: More for awareness than as a suggestion

slactjohnson: Ah, nice; I remember someone mentioning this before, but it fell out of my brain at some point. Thankfully this JSON entry is pretty simple.

scripts/check_archived_pv
ZryletTC: Yeah, there's an unfortunate number of conventions used and I'd be fine with camelCase, lowerwords, or lower_words. I had no reason behind picking archiveStatus, I think I'm just used to it as I've been looking at camViewer a lot recently.
In hindsight, I think I kinda prefer lowerwords for scripts/commands but it really doesn't matter. Maybe at some point we'll converge on one convention but it'll be a painful transition.

silkenelson: I'll own camViewer and the only reason is that I wanted to avoid the same name as the underlying command. Could have gone with cam_viewer......

slactjohnson: Sooooo..... `archive_status` ? `archivestatus`? :)

PR #95, opened 2022-01-19
Changes to ioctool

spenc333:
## Description

Changes to ioctool script involved making searching through config files better, adding the option to open other useful iocInfo files, allowing camera name input to find information, addition of a "cat" option and fixed issues that dealt with multiple iocs.

## Motivation and Context

Fixes #79 .

## How Has This Been Tested?

Tests involved sourcing the script from my epics-dev directory using various camera, pv and ioc names.

## Where Has This Been Documented?

Both scripts have updated usage functions

PR #116, opened 2022-03-23
ENH: Allow serverStat PV input from anywhere

ZryletTC:
## Description

- Primarily changed serverStat's `host_from_PV` function so that it could be used from other hutches or no hutch. Done by calling on ioctool's new `host` option
- Added a `host` option to `ioctool`, reorganized some duplicated code, and added error message for unknown command
- Also changed `PATH` variable setting in all scripts so that current version of engineering_tools is used over other versions that might be in a user's `PATH`

## Motivation and Context

Resolves [LCLSECSD-296](https://jira.slac.stanford.edu/browse/LCLSECSD-296)

## How Has This Been Tested?

Interactively. I believe I've tested all typical cases, but welcome additional testing.

N/A

### DIFF
--- a/scripts/get_curr_exp
+++ b/scripts/get_curr_exp
@@ -32,9 +32,8 @@ do
 	esac
 done
 
-DIRTMP=`dirname  "${BASH_SOURCE[0]}"`
-DIR="$( cd $DIRTMP && pwd -P )"
-PATH=$PATH:$DIR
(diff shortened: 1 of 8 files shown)


ZryletTC:
If #95 is completed, support for camera names could be added pretty easily as well

ZryletTC:
Changing to draft because I want #95 merged first, and there will inevitably be merge conflicts.

PR #117, opened 2022-04-06
camViewer start & stop acquisition

spenc333:
## Description

Added the options to start and stop image acquisition from the command line.

## Motivation and Context

This resolves jira ticket [[LCLSECSD-381]([LCLSECSD-381](https://jira.slac.stanford.edu/browse/LCLSECSD-381))]

## How Has This Been Tested?

Tested from a hutch machine using an established gigE camera

Updated usage function

### DIFF
--- a/scripts/camViewer
+++ b/scripts/camViewer
@@ -158,7 +158,8 @@ OPTIONS:
 -H|--hutch <hutch>  : use a specific hutches camviewer config file
 -e|--enable         : enable camera ioc
 -d|--disable        : disable camera ioc
-
+-a|--aquire         : start acquiring images
+-s|--stop           : stop acquiring images
 EOF
 }
 
@@ -179,9 +180,11 @@ REBOOT=0
 ENABLE=0
 DISABLE=0
 MAINSCREEN=0
+ACQUIRE=0
+STOP=0
 
 
-OPTIONS=$(getopt -o lmredc:w:u:H: --long list,main,reboot,enable,disable,cam:,wait:,rate:,hutch: -n \'$0\' -- "$@")
+OPTIONS=$(getopt -o lmredasc:w:u:H: --long list,main,reboot,enable,disable,acquire,stop,cam:,wait:,rate:,hutch: -n \'$0\' -- "$@")
 
 if [ $? != 0 ] ; then echo "Terminating..." >&2 ; exit 1 ; fi
 
@@ -224,6 +227,14 @@ do
             ;;
         -d|--disable)
             DISABLE=1
+            shift
+            ;;
+        -a|--acquire)
+            ACQUIRE=1
+            shift
+            ;;
+        -s|--stop)
+            STOP=1
             shift
             ;;
         ?)
@@ -335,6 +346,19 @@ if [ $DISABLE -gt 0 ]; then
     imgr "$ioc" --hutch "$iochutch" --disable    
     exit
 fi
+
+if [ $ACQUIRE -gt 0 ]; then
+    echo "Acquiring images for ${CAMNAME}"
+    caput -n "$CAMPVFULL":Acquire 1
+    exit
+fi
+
+if [ $STOP -gt 0 ]; then
+    caput -n "$CAMPVFULL":Acquire 0
+    echo "Done acquiring images for ${CAMNAME}"
+    exit
+fi
+    
 
 IS_ACQUIRING=$(caget -t -n "$CAMPVFULL":Acquire_RBV)
 if isNum "$IS_ACQUIRING"; then


ZryletTC (commented):
Is the request to add two separate options to start/stop or one that cycles the acquisition?
The former doesn't seem super helpful to me while the latter would be an easy fix for some of the issues we've been having.

silkenelson:
I'd prefer stop & start: if we e.g. have an imager, we can reduce network traffic by stopping acquisition when we don't use it. Not that we currently do that, but that might make sense. I think I used to have a 'nudge' method that stopped/started the acquisition - the name might have been chosen after having to pick a letter as 'c' was already taken.

ZryletTC:
Sounds good. In that case though, I would like to also request that we add the nudge/cycle option back in as well.

scripts/camViewer
```diff
@@ -226,6 +230,18 @@ do
...
             ;;
+        -a|--acquire)
+            ACQUIRE=1
+            shift
+            ;;
+        -s|--stop)
+            STOP=1
+            shift
+            ;;
+        --cycle)
```
ZryletTC, line 241: Please add a short option too

scripts/camViewer
```diff
@@ -336,6 +352,26 @@ if [ $DISABLE -gt 0 ]; then
     exit
 fi
 
+if [ $ACQUIRE -gt 0 ]; then
+    echo "Acquiring images for ${CAMNAME}"
```
ZryletTC, line 356:
```suggestion
    echo "Starting acquisition for ${CAMNAME}"
```

scripts/camViewer
```diff
@@ -336,6 +352,26 @@ if [ $DISABLE -gt 0 ]; then
...
+    caput -n "$CAMPVFULL":Acquire 1
+    exit
+fi
+
+if [ $STOP -gt 0 ]; then
+    caput -n "$CAMPVFULL":Acquire 0
+    echo "Done acquiring images for ${CAMNAME}"
```
ZryletTC, line 363:
```suggestion
    echo "Stopping acquisition for ${CAMNAME}"
    caput -n "$CAMPVFULL":Acquire 0
```

scripts/camViewer
```diff
@@ -336,6 +352,26 @@ if [ $DISABLE -gt 0 ]; then
...
+    exit
+fi
+
+if [ $CYCLE -gt 0 ]; then
+    caput -n "$CAMPVFULL":Acquire 0
```
ZryletTC, line 368:
```suggestion
    echo "Stopping and starting acquisition for ${CAMNAME}"
    caput -n "$CAMPVFULL":Acquire 0
```

scripts/camViewer
```diff
@@ -264,14 +264,18 @@ if [ $# -ge 1 ]; then
     exit 1
 fi
 
-
-if [ "$hutch" == ""  ]; then
-    hutch=$(get_info --gethutch)
+hutch_check() {
+hutch=$(get_info --gethutch)
```
ZryletTC, line 268: In general, it's not good practice to make functional changes in commits listed as simple message or documentation changes.
That being said, I like that this makes the error more understandable to the user.
A couple comments though:
- Why did you unindent line 268?
- This block doesn't seem like the right place to print that message on 271 since it will get printed out to the user no matter whether they are running a start/stop command or not. Line 270 and 271 should be used in different cases. Maybe solve with an if/else or moving the echo.

ZryletTC:
Going off what Silke said, it would be nice if the long option was changed to 'nudge' or something else starting with an n if going with '-n' as the short option.

scripts/camViewer
spenc333: 1) I don't think it was meant to be indented. I will have changed in the next commits.
2) Will be updated in the next commits.

README.md
```diff
@@ -80,7 +80,7 @@ usage: /reg/g/pcds/engineering_tools/latest/scripts/camViewer options<br/>
     -d|--disable disable camera ioc<br/>
     -a|--acquire start acquiring images<br/>
     -s|--stop stop acquiring images<br/>
-    --cycle cycles acquisition (first stops then starts) <br/>
+    -n|--cycle cycles acquisition (first stops then starts) <br/>
```
ZryletTC, line 83:
```suggestion
    -n|--nudge cycles acquisition (first stops then starts) <br/>
```

scripts/camViewer
```diff
@@ -264,20 +264,26 @@ if [ $# -ge 1 ]; then
     exit 1
 fi
 
+VH="/reg/g/pcds/epics-dev/spencera/engineering_tools/scripts/verify-hutch"
```
ZryletTC, line 267: What's this? It doesn't look used

scripts/camViewer
```diff
@@ -248,16 +264,26 @@ if [ $# -ge 1 ]; then
...
+                exit 1 
+            else
+                echo 'unknown hutch, specify using -H option'
+                usage
+                exit 1
+            fi
+        fi
     fi
-fi
+}
```
ZryletTC, line 284:
```suggestion
}
```

scripts/camViewer
```diff
@@ -248,16 +264,26 @@ if [ $# -ge 1 ]; then
...
+        if [[ "$hutch" == unknown_hutch ]]; then
+            if [[ $NUDGE -gt 0 ]]; then
+                echo 'To start/stop acquisition, you must be on the hutch specific machine'
+                exit 1 
+            else
+                echo 'unknown hutch, specify using -H option'
+                usage
+                exit 1
+            fi
+        fi
```
ZryletTC, line 282: A couple notes on the hutch check:

1. I don't think it needs to be a function if only run once immediately.
2. If there was a hutch passed, it should be checked using the verify-hutch script.
3. I recommend taking the nudge check out of here and maybe moved down to right above the acquire/stop/nudge section, and check the current hutch against the required hutch if possible. This might be complicated for shared cameras though, so just removing it might be easier.

README.md
ZLLentz: I think this suggestion from last year still applies

scripts/camViewer
```diff
@@ -295,17 +302,27 @@ if [ $CAMNUM -eq 0 ]; then
...
+    if [ "$c" == '' ]; then
+        echo "Camera name not found in hutch. To see which cameras you can access from your current machine use the '-l/--list' option."
+        sleep 3
+    elif [ "$c" == "$CAMNAME" ] && [ "$hutch" != "$(get_info --gethutch)" ] ; then
+        echo "To use camera controls, you must be on the same hutch machine as your camera." 
+        echo "Opening read-only main screen..."
+        sleep 3
+    
+    rm -f $tmp
+    fi
```
ZLLentz, line 324: Two things here:
- It looks like this snippit is copy/pasted from the `name_to_pv` function? Is there a reason why that function couldn't have been re-used?
- This isn't mentioned in the PR text, is this just an incidentally bundled fix?

ZLLentz:
I commented the few small things I had input on, overall it's good.

ZLLentz:
I think this should be good to merge, so I'll do it now

PR #118, opened 2022-04-08
use new XTCAV scripts for dark&lasingOff processing

silkenelson:
Moving from no-longer existing scripts to the 'new' XTCAV setup:

https://confluence.slac.stanford.edu/display/PSDM/New+XTCAV+Documentation

## Motivation and Context
the old method to use makepeds for the XTCAV did no longer work.

## How Has This Been Tested?
by running the scripts 'locally' and using the queue

Not really. It just fixes a couple of available options to actually work.

### DIFF
--- a/scripts/makepeds_psana
+++ b/scripts/makepeds_psana
@@ -74,37 +74,65 @@ check_running_jobs()
 
 xtcav_dark()
 {
-    if [[ $RUNLOCAL == 1 ]]; then
-	python /reg/g/xpp/scripts/xtcav_calibration.py --run $RUN --exp $EXP --dark $ARG
-    else
-	echo bsub -q $QUEUE -o $WORKDIR/xtcav_dark_RUN${RUN}_%J.out python /reg/g/xpp/scripts/xtcav_calibration.py --run $RUN --exp $EXP --dark $ARG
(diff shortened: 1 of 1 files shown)


ZLLentz (commented):
I don't know anything about how these scripts and the batch queue work, so I don't know how to give a useful review. No weird stuff, except for the temp script file which I could be easily convinced is necessary.

My stock suggestion is that `shellcheck` is a good tool for identifying potential issues in shell scripts and I think it'll complain a lot about this file's variable handling in general. You're not obligated to use/follow it in all cases.

silkenelson:
While I won't argue that this script could use an overhaul/rewrite, it is crucial for operations, unfortunately needs modifications/extensions pretty often. This task needs someone not only with decent scripting/coding expertise, but also familiar with what this script is trying to do and how it needs to behave. I don't believe there is anyone in OPS that can currently do this as, so either you can find someone in platforms or just tell me how to merge this without an approval.

ZLLentz:
I can merge without approval, I don't mind

ZLLentz:
Let me find the settings for that so you can merge things without approval, I think you should have that permission

ZLLentz:
The merge with out approval permissions are locked behind enabling 2-factor auth, apparently. I'll just merge this now

silkenelson:
Thank you Zach!

PR #119, opened 2022-04-26
Added hutch-specific configfile, changed to use geometry_deploy_constants

mcb64:
## Description
    - Moved the makepeds-specific flags to the top of the help, just so I could see that this matches makepeds_psana.
    - Created a hutch-specific configuration file /reg/g/pcds/pyps/config/$HUTCH/makepeds.cfg, which contains the ADUMAX, ADUMIN, NOISESIGMAX, and NOISESIGMIN settings for each class of detector.  For now, this only exists in xpp.
    - Removed the -g geometry flag.  Instead, added a bash function "deploy_geometry" that will see if the geometry directory exists already, and if not call the geometry_deploy_constants script to deploy the geometry automatically.  This is also given an optional command that will be run as well if we need to deploy geometry.
    - Call deploy_geometry for cspad, epix10k, jungfrau, and rayonix detectors.  In the event of a jungfrau, also call jungfrau_gain_constants.

## Motivation and Context
See https://jira.slac.stanford.edu/browse/LCLSECSD-1004.

## How Has This Been Tested?
I'm sorry, I don't understand the question.

It hasn't been documented yet.

### DIFF
--- a/scripts/makepeds_psana
+++ b/scripts/makepeds_psana
@@ -23,8 +23,6 @@ OPTIONS:
                make pedestals for Jungfrau - 3 run version(default only cspad/EPIX detectors)
         -q|--queue <queue>
                queue for batch submisson
-        -g|--geometry <expname>
-                deploy geometry <needs experiment to take geometry from>
         -D|--xtcav_dark
                 dark run for XTCAV
         -L|--xtcav_lasingoff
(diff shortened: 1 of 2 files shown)


scripts/makepeds_psana
```diff
@@ -580,26 +614,8 @@ HAVE_JUNGFRAU=`grep Jungfrau /tmp/detnames_$EXP\_$CREATE_TIME | wc -l`
...
-	    $CMD
-	    
-            echo 'Deploy the other jungfrau gain constants'
-	    jungfrau_gain_constants -d exp=${EXP}:run=${RUN} -D
-	    echo 'Deployed the other jungfrau gain constants'
- 	done
-	exit
+    for JUNGFRAU in $DETNAMES; do
+	deploy_geometry $JUNGFRAU "jungfrau_gain_constants -d exp=${EXP}:run=${RUN} -D"
     fi
```
silkenelson, line 619:
```suggestion
    done
```

scripts/makepeds_psana
```diff
@@ -740,41 +741,8 @@ if [[ ( $HAVE_EPIX10K -ge 1 ) && ( $WANT_ZYLA -eq 0 ) && ( $WANT_OPAL -eq 0 ) &&
...
-    	        ORGFILE=`tail -n 1 $GEODIR/$SUBDIR/$DNAME/geometry/HISTORY | awk {'print$1'} | sed s/'file:'//`
-	    fi
-	    CMD=`echo calibfile deploy -t geometry -e $EXP -r 0-end -s $DNAME -f $GEODIR$SUBDIR/$DNAME/geometry/$ORGFILE`
-            echo $CMD
-	    $CMD
-	done
-	exit
+    for EPIX10K in $DETNAMES; do
+	deploy_geometry $EPIX10K
     fi
```
silkenelson, line 746:
```suggestion
    done
```

scripts/makepeds_psana
```diff
@@ -135,6 +133,45 @@ xtcav_lasOff()
...
+}
+
+# Args: Detector ID, Additional command to run.
+deploy_geometry()
+{
+    GEOPATH=`calibfile path -e $EXP -s $1 -t geometry | awk '{print $2;}' -`
+    GEODIR=`dirname $GEOPATH`
+    if [ ! -d $GEODIR ]; then
+	echo No geometry file found, deploying.
+	geometry_deploy_constants -e $EXP -r $RUN -d $1 -D -c ./calib
```
silkenelson, line 165: don't hard code -c (which is also wrong as written), default to standard calib-dir. Could add optional directory later.
```suggestion
	geometry_deploy_constants -e $EXP -r $RUN -d $1 -D
```

mcb64:
I have been asked to run this through shellcheck.  At this point, it complains about the source commands (I'm not using -x), and failure of the cd isn't handled.  The rest is a literal orgy of literals... many, *many* changes to how things are quoted.

I'm afraid, actually.  This could be very very broken, because the changes are so extensive.

silkenelson (approved):
I've tested this for the jungfrau, epix10k2m and epix100 (as an example for a calibrun detector)

PR #121, opened 2022-05-27
Enh/dss check

silkenelson:
## Description
Adding a "dss-check" mode which will run a barebones DAQ recording to the <hutch>daq18 experiment. 

## Motivation and Context
I wanted to make a full checkout of the DSS nodes simpler for the (IT) staff involved. As I made edits, I also ran spellcheck and implemented a fair number of the suggestioins.

## How Has This Been Tested?
Tested new scripts from local checkout in XCS.

No real documentation as these are changes to pre-existing scripts. The new option is not for general use and instructions will be written.

### DIFF
--- a/scripts/ami_offline_psana
+++ b/scripts/ami_offline_psana
@@ -18,8 +18,8 @@ if [[ ($1 == "--help") || ($1 == "-h") ]]; then
 	exit 0
 fi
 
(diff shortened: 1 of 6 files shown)


vespos (approved):
Looks good

PR #122, opened 2022-07-07
Enh/makepeds jf4m epix100

silkenelson:
Add gain-file deployment for epix100 
somewhat better makeup's-command suggestion for takepeds

## Motivation and Context
the detector group would like all detectors to report their values in ADUs and the chosen solution relies on deployment of Gail-files in the calib-directory

## How Has This Been Tested?
Tested on xppx40719 and xpplz0620 (gain files removed later).

### DIFF
--- a/scripts/makepeds_psana
+++ b/scripts/makepeds_psana
@@ -173,6 +173,20 @@ deploy_geometry()
     fi
 }
 
+# Args: Detector ID, Additional command to run.
+deploy_gain()
+{
+    GAINPATH=$(calibfile path -e "$EXP" -s "$1" -t pixel_gain | awk '{print $2;}' -)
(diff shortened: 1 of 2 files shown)


vespos (approved):
Looks good overall. Just some question to undrestand things:
1. Is the -F argument deprecated now for makepeds?
2. What is the $CALIBCODE variable and why are the number of event considered NUMEVT increased by a factor 10?
3. What does --upload do for the JF?

silkenelson:
-F still does something when you don't use the queue. How it was written though seems to have made people think you needed both -F and -q <queue>.

$CALIBCODE allows us to select events with a given event code (either select 162 or select NOT 137). I suspect one of the countes counts all events, not only the ones that we select, to the default number needs to be raised.

PR #126, opened 2022-07-20
DEP: Remove gige script and serverStat's isCam function

### LINKED ISSUE #124: Rewrite isCam from serverStat!
## Expected Behavior
serverStat is supposed to be used to cycle servers as a wrapper on psipmi.
It is NOT supposed to fix cameras or tell people to use the iocmanager to reboot servers, if anything, the opposite!

## Current Behavior
XPP report this to happen for serverStat yag2 (DAQ camera!):
when I try to serverStat is it says “go to the ioc manager to reboot/check status” in the ioc manager I find SB3 - YAG 2 and “noconnect”

## Possible Solution
remove the whole isCam section and at the very least, don't have it find things that also actively exist in the DAQ cnf file

ZryletTC:
## Description

- Removes functionality from serverStat that allows it to interact with giges. We have other tools for that and duplicated functionality leads to confusion
- Fixes the dss node grabbing given an event# device name
- Additional cleanup to the serverStat script
- Also removes the deprecated gige script

## Motivation and Context

Closes #124 

## How Has This Been Tested?

Tested with cxi things from cxi and psbuild.
DSS check tested in all HXR hutches.

Here


ZryletTC:
One last issue I want to solve before merge: When I was testing, I noticed the section for event/dss name checking doesn't work and instead uses the full list of ips, which just fails.

ZryletTC:
Ok, I added the dss check like I was hoping and also added an exit when hostnames weren't recognized, instead of giving a bad error message. Tested with cxi things from cxi and psbuild. Should also test with other hutches devices.

ZryletTC:
Tested dss check in all hxr hutches. Ready to merge once approved.

vespos (approved):
LGTM

vespos:
@silkenelson should I just merge that or do you want to review as well?

silkenelson:
@vespos Feel free to go ahead.

PR #130, opened 2022-09-21
Event count by detector type

slacmshankar:
Add a script/report for the detector group. The report generates event count by detector type.


vespos (approved):
Might want to run it through flake8, but overall looks good.

PR #131, opened 2022-10-17
Bug/makepeds epix100

silkenelson:
## Description
add xtc-path for geometry deploy & epix100 gain file to make them also work on the DRPSRCF FFB
also fix usage text for checkcnf.

## Motivation and Context
geometries could not deployed from jobs run against the FFB.

## How Has This Been Tested?
Running the command.

### DIFF
--- a/scripts/checkcnf
+++ b/scripts/checkcnf
@@ -4,7 +4,8 @@ usage()
 cat << EOF
 usage: $0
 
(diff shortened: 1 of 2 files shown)


mcb64 (commented):
Hmm... not sure why the previous review didn't work, but...

mcb64 (approved):
Because I'm an idiot and didn't check the "Approve" radio button.

PR #132, opened 2022-10-19
MAINT: workaround for opengl crash when sshing to operator console

ZLLentz:
## Description

Basic implementation of https://jira.slac.stanford.edu/browse/LCLSPC-477
which is a workaround for https://jira.slac.stanford.edu/browse/LCLSECSD-1446

@JJL772 showed this to us months ago after I'd already hotfixed the local pydm and typhos code to avoid the opengl crashes we were seeing on the operator consoles. See the linked jira tickets for an extended background.

## Motivation and Context

I'd like to go back to running non-modified versions of pydm and typhos. An environment variable is by far the most preferable workaround so that we can run out of the box software.

There's something esoteric for IT to fix about this whole situation, but for now let's at least make sure python apps never crash by this bug again so it becomes somebody else's problem.

## How Has This Been Tested?

Interactively only

Comments only

### DIFF
--- a/scripts/pcds_conda
+++ b/scripts/pcds_conda
@@ -128,7 +128,8 @@ source "${TOKEN_DIR}/typhos.sh"
 # The same was done with PyDM and this fixes Designer and friends.
 if [ -n "${SSH_CONNECTION}" ]; then
 	export QT_QUICK_BACKEND="software"
-  # On operator consoles over ssh for most users, opengl crashes unless this
+  # On operator consoles over ssh for most users, opengl crashes unless this is set.
+  # This tells the GLX extension to use software rendering by way of the mesa library.
   if [ -z "${__GLX_VENDOR_LIBRARY_NAME}" ]; then
     export __GLX_VENDOR_LIBRARY_NAME="mesa"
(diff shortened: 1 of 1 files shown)


JJL772 (approved):

klauer (approved):

tangkong (approved):

scripts/pcds_conda
```diff
@@ -128,4 +128,8 @@ source "${TOKEN_DIR}/typhos.sh"
 # The same was done with PyDM and this fixes Designer and friends.
 if [ -n "${SSH_CONNECTION}" ]; then
 	export QT_QUICK_BACKEND="software"
+  # On operator consoles over ssh for most users, opengl crashes unless this
```
ZLLentz, line 131: Duplicating Ken's docs suggestion to the other file
Yes I know it's pretty bad that these files are duplicated like this maybe that's another thing to fix
```suggestion
  # On operator consoles over ssh for most users, opengl crashes unless this is set.
  # This tells the GLX extension to use software rendering by way of the mesa library.
```

PR #133, opened 2022-10-26
FIX: pcds_conda, dev_conda: only set mesa if renderer is failing

ZLLentz:
## Description

Check if the server has a problem before switching the renderer library.
This is WIP because it is messy and littered with echo and I'm not sure if it's even fully right.

## Motivation and Context

Setting the `__GLX_VENDOR_LIBRARY_NAME` didn't work for all consoles.
For many cases it actually created a problem rather than solving it.

Only do it if we definitely fail without it.

That is, when:
- glxinfo has a nonzero exit code (usually 0 is success, 1 is errors, 139 is segfault)
- glxinfo has a 0 exit code but shows errors in the text (TODO: does this actually happen or is it a remnant of an older version of the script?)

## How Has This Been Tested?

Heavily interactively.

N/A

### DIFF
--- a/scripts/dev_conda
+++ b/scripts/dev_conda
@@ -53,11 +53,35 @@ source "${TOKEN_DIR}/typhos.sh"
 # Revert to Software Raster when SSH to avoid issues with QtQuick.
 # The same was done with PyDM and this fixes Designer and friends.
 if [ -n "${SSH_CONNECTION}" ]; then
+  echo "SSH Session"
 	export QT_QUICK_BACKEND="software"
-  # On operator consoles over ssh for most users, opengl crashes unless this is set.
-  # This tells the GLX extension to use software rendering by way of the mesa library.
-  if [ -z "${__GLX_VENDOR_LIBRARY_NAME}" ]; then
-    export __GLX_VENDOR_LIBRARY_NAME="mesa"
+  # Check for errors in the renderer
+  # Manually fall back to mesa library if renderer doesn't work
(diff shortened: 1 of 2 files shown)


ZLLentz:
Anecdotally, I think I never have hit the warning trigger, nor could I ever make it happen again in testing now.
Previously I wasn't getting the return code out of the glxinfo command properly due to some things I was doing wrong, so I was parsing for the error text instead. But since I'm not messing that up any more it should suffice to just check the error code to see if there is a problem or not.

So I will:
- simplify the script a bit
- remove echo
- test again using xpp

ZLLentz:
I think this is simplified to the important bits now, I still need to waltz to XPP for another check

ZLLentz:
I've double checked that this works for my laptop to any -control, for tmo-control -> tmo-daq, for xpp psusr -> xpp-control, and from rix-console -> rix-montitor

I think the more concise expression in this version is pretty good

tangkong (approved):
Seems good to me.  While I wonder if we could avoid copy-pasting the same script snippet in two conda inits, that's probably an issue for a later time.

The core of this check has worked for me throughout, but I've never been impacted particularly heavily by this... somehow. 🤷

ZLLentz:
I'm going to throw this into live latest for engineering tools to eliminate the echo statements and etc. and wait for Ken's approval too in case there's some way he sees that this can be cleaner

ZLLentz:
Goal is to merge and tag engineering tools again afterwards

PR #135, opened 2022-10-28
ENH/Add SmarAct MSC2 to motor-expert-screen

vespos:
## Description
Add the right expert screen for SmarAct stages

## Motivation and Context
Have the expert screen script open the right screen for SmarAct

## How Has This Been Tested?
Tested with some axis of the LIB SmarAct IOC at XPP.

Docstring in the script


silkenelson (approved):
Looks good to me.

PR #137, opened 2022-11-01
fix makepeds geometry and gain deployment after backend change

vespos:
## Description
Fix makepeds gain and geometry deployment

## Motivation and Context
Backend changed messed up the deployment of files from the ffb

## How Has This Been Tested?
xppx1004421 run 1

N/A

### DIFF
--- a/scripts/makepeds_psana
+++ b/scripts/makepeds_psana
@@ -162,10 +162,9 @@ deploy_geometry()
     GEODIR=$(dirname "$GEOPATH")
     if [ ! -d "$GEODIR" ]; then
 	echo No geometry file found, deploying.
-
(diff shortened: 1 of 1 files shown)


vespos:
Need to fix the part with the deployment of the gain for the jungfrau. This probably has to be done at the level of the jungfrau_gain_constants function.

silkenelson:
I don't really want to approve this because I feel pretty strongly that the -x argument should take the _same_ information for gain & geometry as it does for pedestal creation.

vespos:
Yes, let's keep it open until we settle on something with Mikhail.

vespos:
I think we can abandon this PR, given that MIkhail has improved things on his side?

silkenelson:
Agreed.

PR #138, opened 2022-11-01
MAINT: cleaner handling of latest conda and dev conda

### LINKED ISSUE #136: pcds_conda drops you into the new env even before it's ready
## Expected Behavior

`source pcds_conda` without the env variable set should drop you into the latest environment that exists and is fully-formed.

## Current Behavior

If I've pressed the "deploy env" button but the environment isn't complete, it still drops you into the new environment.

## Possible Solution

Maybe there's an automatic way to know if the env is ready?
Or maybe it's better and safer to hard-code the "latest" env name and update it manually once ready.

ZLLentz:
## Description

- For `dev_conda`, use the latest tag as a base instead of an arbitrary outdated `dev` environment. This gives us a better chance of dev always being up-to-date and stable.
- For `pcds_conda` and `dev_conda`, find the latest tag via config file instead of automatically.

## Motivation and Context

The outdated `dev` environment is causing issues because the base packages are so far behind and it's not possible to update it safely in-place.
The dynamically picked `pcds_conda` environment path causes issues (closes #136)
Now, we will be able to test and vet the fully installed environment before dropping people into it by default.

I've placed the config file for this inside of the conda environment itself. This keeps it one level hidden from accidental changes and allows a different config seamlessly when we boost to the next version of Python.

See `/cds/group/pcds/pyps/conda/py39/latest_env`

## How Has This Been Tested?

Interactively only.

N/A, should be transparent to users.

### DIFF
--- a/scripts/dev_conda
+++ b/scripts/dev_conda
@@ -28,8 +28,8 @@ source "${BASE_ENV}/etc/profile.d/conda.sh"
 
 # Activate base env first so things like conda build are available
 conda activate
-# Default to dev pcds release
-conda activate dev
+# Default to latest pcds release as the base
+conda activate "$(cat "${BASE_ENV}/latest_env")"
 
 PCDSDEVICES_DEV_UI="${APPS_ROOT}/dev/pcdsdevices/pcdsdevices/ui"
 
--- a/scripts/pcds_conda
+++ b/scripts/pcds_conda
@@ -65,9 +65,7 @@ if [[ -z "${PCDS_CONDA_VER}" ]]; then
   # If env var unset, then default to latest
   BASE_ENV="${CONDA_ROOT}/py39"
   PYTHON_VER="3.9"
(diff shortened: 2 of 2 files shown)


ZLLentz:
This is not to be deployed prior to the pamm in case the updated dev causes unforeseen and esoteric issues.

tangkong (approved):
I'm in favor of this.  dev has been treated as simply another tag, with some apps (lucid-launcher) being based on it.  If users want a frozen env they should probably just pick a tag.

klauer (approved):
This looks good to me. 

 I like that the `latest_env` is something we can just set in a separate file, and roll back to if need be.
"Patch in development packages on top of the latest env" is a very sensible approach for our development conda environment.

ZLLentz:
Great! I'll merge and send out an additional engineering_tools tag during the POMM a week from today.

PR #139, opened 2022-11-01
FIX: pydev_env was not showing dev typhos templates

ZLLentz:
## Description

Unset the long `PYDM_DISPLAYS_PATH` from `pcds_conda` in the `pydev_env` setup

## Motivation and Context

The path here was masking personal dev env screens with the production screens, preventing testing of dev typhos templates when using `pydev_env`.

## How Has This Been Tested?

Interactively only

Release notes later

### DIFF
--- a/scripts/pydev_env
+++ b/scripts/pydev_env
@@ -18,3 +18,5 @@ source pcds_conda
 PYDEV="${HOME}/pydev"
 export PATH="${PYDEV}/bin:${PATH}"
 export PYTHONPATH="${PYDEV}:${PYTHONPATH}"
+# The following is set by pcds_conda and needs to be undone for dev typhos to pick your templates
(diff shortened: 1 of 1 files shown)


tangkong (approved):
lgtm 👍

klauer (approved):
This looks good to me.

A slight tangen and a potential topic for a typhos issue:
I ran into (I think) the same thing maybe last month - I recall seeing a confusing set of `DeviceName.ui` listed twice in typhos due to it finding the display once in the production installation (first) and once in the development install (second):
image
Perhaps it'd be worthwhile to list a more complete filename in that list when two have an identical name...

ZLLentz:
I saw exactly the same thing in testing but I didn't realize what it was. Good catch.

PR #140, opened 2022-11-03
Fix/makeped jf fraclm

silkenelson:
## Description
Adding a parameter for the jungfrau pedestal status to makepeds(_psana)

## Motivation and Context
the default in the current release ana-4.0.47-py3 flags about half the jungfrau1M pixels as bad....bad means that 1 event in a run has a very low or high pedestal value.

## How Has This Been Tested?
Running & checking the status...

### DIFF
--- a/scripts/makepeds_psana
+++ b/scripts/makepeds_psana
@@ -198,7 +198,6 @@ deploy_gain()
     fi
 }
 
-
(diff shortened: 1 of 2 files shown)


vespos (approved):
LGTM

PR #141, opened 2022-12-13
Check for failure before declaring success.

mcb64:
makepeds declares success after calling the geometry or gain deployment scripts.  This should be conditional on the return code of the script.

Declaring failure if the script fails is much less confusing.


silkenelson (approved):
Looks good to me.

PR #142, opened 2022-12-14
ENH: Added check for Beckhoff axis

jortiz-slac:
Added a check for Beckhoff motors and launch the typhos motor screen. 

## Motivation and Context

Previously, Beckhoff motors defaulted to the Aerotech edm screen.

## How Has This Been Tested?

Tested the script by calling IMS motors and PLC motors. Ensured that the typhos window title iscorrectly defined so the repositioning works as intended.

Added comments where the check is performed.

### DIFF
--- a/scripts/motor-expert-screen
+++ b/scripts/motor-expert-screen
@@ -1,5 +1,3 @@
-isbeckhoff=false
-
 #!/bin/bash
 
 OLD_VERSION='R2.5.0'
--- a/scripts/motor-expert-screen
+++ b/scripts/motor-expert-screen
@@ -31,10 +31,21 @@ export EDMDATAFILES=.:${PCDS_EDMS}/xps8:/reg/g/pcds/package/epics/3.14/ioc/commo
 source /reg/g/pcds/setup/pathmunge.sh
 ldpathmunge /usr/local/lib
 
+# Check to see if this is a Beckhoff axis
+#   caget will return exit code 0 if :PLC:nErrorId_RBV exists, 
+#   we set isbeckhoff based on this.
+isbeckhoff=false
+caget "${PREFIX}:PLC:nErrorId_RBV" > /dev/null 2>&1
+if [ $? -eq 0 ]; then
+    isbeckhoff=true
+fi
+
 # Choose title so we can move the right window.
 # If this script hangs later, titles probably changed.
 rtyp=`caget -t $PREFIX.RTYP` > /dev/null 2>&1
-if [ "${rtyp}" == 'xps8p' ]; then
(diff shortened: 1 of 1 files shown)


scripts/motor-expert-screen
```diff
@@ -31,10 +31,21 @@ export EDMDATAFILES=.:${PCDS_EDMS}/xps8:/reg/g/pcds/package/epics/3.14/ioc/commo
 source /reg/g/pcds/setup/pathmunge.sh
 ldpathmunge /usr/local/lib
 
+# Check to see if this is a Beckhoff axis
+#   caget will return exit code 0 if :PLC:nErrorId_RBV exists, 
+#   we set isbeckhoff based on this.
+isbeckhoff=false
+caget "${PREFIX}:PLC:nErrorId_RBV" > /dev/null 2>&1
```
ZLLentz, line 38: This is an extra 1 second timeout on all non-beckhoff axes. Is it possible to do these checks inside the rtyp == motor branch so that there is minimal impact on other kinds of motors?

scripts/motor-expert-screen
```diff
@@ -31,10 +31,21 @@ export EDMDATAFILES=.:${PCDS_EDMS}/xps8:/reg/g/pcds/package/epics/3.14/ioc/commo
...
+if [ $? -eq 0 ]; then
+    isbeckhoff=true
+fi
+
 # Choose title so we can move the right window.
 # If this script hangs later, titles probably changed.
 rtyp=`caget -t $PREFIX.RTYP` > /dev/null 2>&1
-if [ "${rtyp}" == 'xps8p' ]; then
+if [ ${isbeckhoff} == true ]; then
+    title="Typhos Suite - ${PREFIX}"
```
ZLLentz, line 47: I think this is unused

ZLLentz (commented):
Pretty good approach to doing this, just a couple comments

scripts/motor-expert-screen
jortiz-slac: Yeah, sure. I originally put it outside to keep the nesting as shallow as possible but the one second timeout is not great. I will update.

scripts/motor-expert-screen
jortiz-slac: The title is used to reposition the new window under the mouse.

ZLLentz: Ah, gotcha! Somehow I thought this was being applied, rather it is set to match the truth.

ZLLentz (approved):
Seems good to me

vespos (approved):

ZLLentz:
I'll go ahead and merge this in

PR #144, opened 2023-01-19
Fully test the event masks and fix temp file name

mcb64:
Added the PID to the temporary file name to hopefully avoid collisions should two people be running the script at once.

The old version of the script only tested a few possible event mask values, the largest being 16, which didn't seem to be correct anyway.  The detail part of the script has been rewritten to really test the values in the event mask, and also complain if the mask contains any values *not* supported by an event process.


silkenelson (approved):
LGTM

PR #145, opened 2023-02-03
WIP/ENH: adds xpp mono-pink line switchover script

tangkong:
## Description
Creates a script to switch relevant XPP devices between the mono (L2) and pink (L0) branches depending on the insertion state.  Looks for devices inside a z range on either the l2 or l0 branch

## Usage

``` console
bash $ xpp_update_happi_line -h

usage: xpp_update_happi_line [-h] [--min MIN_Z] [--max MAX_Z] [--dry-run]

Updates items in the happi database to reflect the which line XPP is using (mono line, pink line).
Assumes pcds-conda, and grabs the currently active happi config.

optional arguments:
  -h, --help   show this help message and exit
  --min MIN_Z  min z range
  --max MAX_Z  max z range
  --dry-run    run script without saving to database
```

## Motivation and Context
XPP currently has a group of devices that gets switched between the L2 and L0 branches.  In order for lightpath to properly represent this in every hutch's lightpath, the happi database must reflect this change.  

Note: for lightpath to include the L2 branch in the facility graph, the L0 -> (L0, L2) branching device must be properly reporting its lightpath status (xpp_lodcm)

## How Has This Been Tested?
Interactively for now

[JIRA](https://jira.slac.stanford.edu/browse/LCLSECSD-1630)


scripts/xpp_update_happi_line
```diff
@@ -0,0 +1,88 @@
...
+
+    parser.add_argument('--dry-run', action='store_true', dest='dry_run',
+                        default=False, required=False,
+                        help='run script without saving to database')
+
+    args = parser.parse_args()
+    max_z = args.max_z
+    min_z = args.min_z
+    pink_active = EpicsSignalRO('XPP:INS:POS:01:IN_DI').get()
+    mono_active = EpicsSignalRO('XPP:INS:POS:01:OUT_DO').get()
```
ZLLentz, line 35: I think this is a good approach. The only other approach I can think of is having a mechanism for creating happi devices that "move/modify" other happi devices, so then the xpp table can be one such device. That might get complicated quickly and not be worth the hassle.

vespos:
Just tested in both in dry-run and for real, it's working great!
I just added the possibility to force the change to either lines, this makes it easier to test things and we might need it at some point.

scripts/xpp_update_happi_line
```diff
@@ -0,0 +1,102 @@
...
+    l2_results = client.search_range('z', min_z, max_z, input_branches=['L2'],
+                                     output_branches=['L2'])
+
+    # assess current state
+    print(f'Updating XPP mono-pink line devices z=({min_z}, {max_z})...')
+    if not pink_active and mono_active:
+        print('  Detected mono line active (L2, sharing beam)')
+        new_branch = ['L2']
+        results = l0_results
+        if not (len(l0_results) > 0):
```
klauer, line 65: Nitpick - saves typing and is easier to read:
```suggestion
        if not l0_results:
```

scripts/xpp_update_happi_line
```diff
@@ -0,0 +1,102 @@
...
+            print('xx no devices on L0 branch to transition')
+            return
+    elif not mono_active and pink_active:
+        print('  Detected pink line active (L0, potentially blocking beam)')
+        new_branch = ['L0']
+        results = l2_results
+        if not (len(l2_results) > 0):
```
klauer, line 72: Same here

scripts/xpp_update_happi_line
```diff
@@ -0,0 +1,102 @@
+#!/usr/bin/env python
+"""
+Updates items in the happi database to reflect the which line XPP is 
+using (mono line, pink line).
+
+Assumes pcds-conda, and grabs the currently active happi config.
```
klauer, line 6: This is probably good enough for this little script. Just to note, some other scripts sometimes have a Python source code file along with a bash entrypoint that make the running environment consistent:
```bash
source pcds_env
cd $SCRIPT_PATH
python the_script.py "$@"
```

klauer (approved):
I think this looks good 👍 

(It's also maybe the first custom happi-mass-editing tool?)

scripts/xpp_update_happi_line
tangkong: good thought, I can add this.

scripts/xpp_update_happi_line
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
+source pcds_conda
+cd $ENG_TOOL_SCRIPTS
```
ZLLentz, line 3: Where does this environment variable get set? I don't have it in my terminal.

scripts/xpp_update_happi_line.py
```diff
@@ -0,0 +1,102 @@
...
+import happi
+from ophyd import EpicsSignalRO
+
+
+def main():
+    parser = argparse.ArgumentParser(
+        prog='xpp_update_happi_line', description=__doc__
+    )
+
+    group = parser.add_mutually_exclusive_group()
```
ZLLentz, line 21: I wish I knew about `add_mutually_exclusive_group` before some CLI stuff I did recently

ZLLentz (approved):
One small question because I can't run the script myself
Otherwise looks good

scripts/xpp_update_happi_line.py
tangkong: Thanks to @vespos for this one

scripts/xpp_update_happi_line
tangkong: For some reason I have this sourced, finding exactly where from is tough as always

https://github.com/pcdshub/epics-config/blob/1403e2c62876d4c0b469e687816b569ef09da4fe/common_dirs.sh#L32

ZLLentz: Huh, interesting. Note then that the wrapper here will always run the latest version of the python script that's been updated in that specific folder, unless you set it again yourself.

PR #146, opened 2023-02-24
Bare repo script

wwright-slac:
There are a lot of IOCs that are not properly version controlled since we moved from svn to git. Often, I run into IOCs that do not have a hutch specific bare repo setup, so I made this script to help facilitate the process of creating that bare repo and the subsequent steps afterwards.

This script: 

1. Initializes the local repository
2. Checks in the .cfg file, Makefile and .gitignore
3. Creates the bare repo in the proper location using eco_tools git-bare-repo script
4. Sets this newly created bare repo as the remote origin and sets it as the upstream

## Motivation and Context
As mentioned above

## How Has This Been Tested?
This has been tested once with the Smarpod IOC, which I had yet to create a hutch bare repo for. Worked perfectly. 

Code includes simple explanations for how to use the script.

If this script is redundant, or could be useful in another script or repository, feel free to shoot this down.

### DIFF
--- a/scripts/new_bare_repo
+++ b/scripts/new_bare_repo
@@ -37,6 +37,8 @@ then
     git-bare-repo /afs/slac.stanford.edu/g/cd/swe/git/repos/package/epics/ioc/${HUTCH}/${REPO}.git
     git remote add origin /afs/slac.stanford.edu/g/cd/swe/git/repos/package/epics/ioc/${HUTCH}/${REPO}.git
     git push -u origin master
-else
-    echo "/afs/slac.stanford.edu/g/cd/swe/git/repos/package/epics/ioc/${HUTCH}/ does not exist"
+elif [ ! -d "/afs/slac.stanford.edu/g/cd/swe/git/repos/package/epics/ioc/${HUTCH}/" ]
+    echo "The directory /afs/slac.stanford.edu/g/cd/swe/git/repos/package/epics/ioc/${HUTCH}/ does not exist!"
+elif [ -d "/afs/slac.stanford.edu/g/cd/swe/git/repos/package/epics/ioc/${HUTCH}/${REPO}.git" ]
+    echo "There already exists a bare repo for this device!"
 fi
--- a/scripts/new_bare_repo
+++ b/scripts/new_bare_repo
@@ -15,20 +15,25 @@ You also might need to run kinit and aklog before hand!
 EOF
 }
 
-if [ $# -lt 1 ]; then
+if [[ ($1 == "--help") || ($1 == "-h") ]]; then
+   usage
+   exit 0
+
+elif [ $# -lt 2 ]; then
    echo 'Need to specify the hutch and name of the proposed bare repo' >&2
    usage
    exit 1
 
-elif [[ ($1 == "--help") || ($1 == "-h") ]]; then
-   usage
-   exit 0
 fi
 
 HUTCH=$1
 REPO=$2
+CWD=$(PWD)
 
-if [ -d "/afs/slac.stanford.edu/g/cd/swe/git/repos/package/epics/ioc/${HUTCH}/" ] && [ ! -d "/afs/slac.stanford.edu/g/cd/swe/git/repos/package/epics/ioc/${HUTCH}/${REPO}.git" ]
+Check if current working directory is 
+
+
+if [ -d "/afs/slac.stanford.edu/g/cd/swe/git/repos/package/epics/ioc/${HUTCH}/" ] && [ ! -d "/afs/slac.stanford.edu/g/cd/swe/git/repos/package/epics/ioc/${HUTCH}/${REPO}.git" ] && [${CWD} == *${REPO}* ]
 then
     git init
     cp /cds/sw/tools/bin/eco_tools/gitignore.template .gitignore
@@ -39,6 +44,8 @@ then
     git push -u origin master
 elif [ ! -d "/afs/slac.stanford.edu/g/cd/swe/git/repos/package/epics/ioc/${HUTCH}/" ]
     echo "The directory /afs/slac.stanford.edu/g/cd/swe/git/repos/package/epics/ioc/${HUTCH}/ does not exist!"
+    exit 1
 elif [ -d "/afs/slac.stanford.edu/g/cd/swe/git/repos/package/epics/ioc/${HUTCH}/${REPO}.git" ]
     echo "There already exists a bare repo for this device!"
+    exit 1
 fi
--- a/scripts/new_bare_repo
+++ b/scripts/new_bare_repo
@@ -1,0 +1,42 @@
+#!/bin/bash
+usage()
+{
+cat << EOF
+usage: new_bare_repo <hutch> <repo name>
+
+Script for automating the tedious process of integrating an existing IOC into our git bare repo scheme
+Run in the top level of your IOC
+
+Ex. new_bare_repo rix new_device
+(Creates a new bare repo in /afs/slac.stanford.edu/g/cd/swe/git/repos/package/epics/ioc/rix/ named "new_device.git")
+
+You also might need to run kinit and aklog before hand!
+
+EOF
+}
+
+if [ $# -lt 1 ]; then
+   echo 'Need to specify the hutch and name of the proposed bare repo' >&2
+   usage
+   exit 1
+
+elif [[ ($1 == "--help") || ($1 == "-h") ]]; then
+   usage
+   exit 0
+fi
+
+HUTCH=$1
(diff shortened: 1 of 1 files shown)


scripts/new_bare_repo
```diff
@@ -0,0 +1,44 @@
...
+cat << EOF
+usage: new_bare_repo <hutch> <repo name>
+
+Script for automating the tedious process of integrating an existing IOC into our git bare repo scheme
+Run in the top level of your IOC
+
+Ex. new_bare_repo rix new_device
+(Creates a new bare repo in /afs/slac.stanford.edu/g/cd/swe/git/repos/package/epics/ioc/rix/ named "new_device.git")
+
+You also might need to run kinit and aklog before hand!
```
ZLLentz, line 13: Out of scope for this PR, I'm thinking we might want to add an engineering tools script that checks if kinit/aklog needs to be run (and optionally, runs it). This is doable using the output of the `klist` command.

Would something like this have been useful when writing this script?

scripts/new_bare_repo
```diff
@@ -0,0 +1,44 @@
...
+
+EOF
+}
+
+if [ $# -lt 1 ]; then
```
ZLLentz, line 18: Should this be a 2? Or does this count differently than I think it does? I'm expect as-written that the script would let you run with a blank repo which might behave oddly.

I think the help check needs to become the first part of this if/else and this needs to be boosted to 2

scripts/new_bare_repo
```diff
@@ -0,0 +1,44 @@
...
+then
+    git init
+    cp /cds/sw/tools/bin/eco_tools/gitignore.template .gitignore
+    git add *.cfg Makefile .gitignore
+    git commit -m "Initial commit"
+    git-bare-repo /afs/slac.stanford.edu/g/cd/swe/git/repos/package/epics/ioc/${HUTCH}/${REPO}.git
+    git remote add origin /afs/slac.stanford.edu/g/cd/swe/git/repos/package/epics/ioc/${HUTCH}/${REPO}.git
+    git push -u origin master
+elif [ ! -d "/afs/slac.stanford.edu/g/cd/swe/git/repos/package/epics/ioc/${HUTCH}/" ]
+    echo "The directory /afs/slac.stanford.edu/g/cd/swe/git/repos/package/epics/ioc/${HUTCH}/ does not exist!"
```
ZLLentz, line 41: Should this elif branch end in `exit 1`?

scripts/new_bare_repo
```diff
@@ -0,0 +1,44 @@
...
+   usage
+   exit 0
+fi
+
+HUTCH=$1
+REPO=$2
+
+if [ -d "/afs/slac.stanford.edu/g/cd/swe/git/repos/package/epics/ioc/${HUTCH}/" ] && [ ! -d "/afs/slac.stanford.edu/g/cd/swe/git/repos/package/epics/ioc/${HUTCH}/${REPO}.git" ]
+then
+    git init
```
ZLLentz, line 33: Out of scope (?): a future version of this script could double-check that we're not running `git init` in a random directory (some sort of path matching against ${REPO}?)

scripts/new_bare_repo
```diff
@@ -0,0 +1,44 @@
...
+    cp /cds/sw/tools/bin/eco_tools/gitignore.template .gitignore
+    git add *.cfg Makefile .gitignore
+    git commit -m "Initial commit"
+    git-bare-repo /afs/slac.stanford.edu/g/cd/swe/git/repos/package/epics/ioc/${HUTCH}/${REPO}.git
+    git remote add origin /afs/slac.stanford.edu/g/cd/swe/git/repos/package/epics/ioc/${HUTCH}/${REPO}.git
+    git push -u origin master
```
ZLLentz, line 39: Out of scope: a future version of this script could help set up a github repo instead of an afs repo.
At some ill-defined future point we were going to integrate code review into our IOC process.

ZLLentz (commented):
I have two concrete suggestions and some musing for future follow-ups

scripts/new_bare_repo
wwright-slac: No, you are reading that correctly, that should be 2

scripts/new_bare_repo
wwright-slac: Hmm... The issue is that peoples dev directories, where they are likely building these IOCs, are unlikely to follow the same kind of path structure. It would be easy to add something that looks for ${REPO} _somewhere_ in the path, but that wouldn't cover all possible cases

scripts/new_bare_repo
wwright-slac: Personally, I am not even sure when to run kinit or not... It's usually something I only run when I get an error message with git that I can't explain otherwise

scripts/new_bare_repo
```diff
@@ -0,0 +1,49 @@
...
+elif [ $# -lt 2 ]; then
+   echo 'Need to specify the hutch and name of the proposed bare repo' >&2
+   usage
+   exit 1
+
+fi
+
+HUTCH=$1
+REPO=$2
+CWD=$(PWD)
```
klauer, line 31: It's either $PWD (as in the env var) or `$(pwd)` as in "run [pwd](https://man7.org/linux/man-pages/man1/pwd.1.html) and get its output"
```suggestion
CWD=${PWD}
```

scripts/new_bare_repo
```diff
@@ -0,0 +1,49 @@
...
+then
+    git init
+    cp /cds/sw/tools/bin/eco_tools/gitignore.template .gitignore
+    git add *.cfg Makefile .gitignore
+    git commit -m "Initial commit"
+    git-bare-repo /afs/slac.stanford.edu/g/cd/swe/git/repos/package/epics/ioc/${HUTCH}/${REPO}.git
+    git remote add origin /afs/slac.stanford.edu/g/cd/swe/git/repos/package/epics/ioc/${HUTCH}/${REPO}.git
+    git push -u origin master
+elif [ ! -d "/afs/slac.stanford.edu/g/cd/swe/git/repos/package/epics/ioc/${HUTCH}/" ]
+    echo "The directory /afs/slac.stanford.edu/g/cd/swe/git/repos/package/epics/ioc/${HUTCH}/ does not exist!"
```
klauer, line 44: A general principle I like to follow is that of "exiting early" (i.e., when conditions aren't right).

That is to say, I think it would make more sense to check all directories at the start of the script before doing the bare repo creation.

```
# check if hutch directory doesn't exist (and exit if so)
# check if repo directory already exists (and exit if so)
# check if "REPO" not in $PWD (and exit if so)
# otherwise, do the thing
```

scripts/new_bare_repo
```diff
@@ -0,0 +1,49 @@
...
+   exit 1
+
+fi
+
+HUTCH=$1
+REPO=$2
+CWD=$(PWD)
+
+
+if [ -d "/afs/slac.stanford.edu/g/cd/swe/git/repos/package/epics/ioc/${HUTCH}/" ] && [ ! -d "/afs/slac.stanford.edu/g/cd/swe/git/repos/package/epics/ioc/${HUTCH}/${REPO}.git" ] && [${CWD} == *${REPO}* ]
```
klauer, line 34: Whitespace matters around `[` and `]` since `[` is actually a program (see `which [` or try `if [a==a]; then echo yes; fi` vs `if [ a == a ]; then echo yes; fi`
```suggestion
if [ -d "/afs/slac.stanford.edu/g/cd/swe/git/repos/package/epics/ioc/${HUTCH}/" ] && [ ! -d "/afs/slac.stanford.edu/g/cd/swe/git/repos/package/epics/ioc/${HUTCH}/${REPO}.git" ] && [ ${CWD} == *${REPO}* ]
```

scripts/new_bare_repo
klauer: Might be good to leave this `git push` step be manual, just in case the script does something you don't expect

scripts/new_bare_repo
ZLLentz: It needs some thought for sure, but I think it would be awkward to accidentally/automatically run git init somewhere in the middle of the file tree

scripts/new_bare_repo
ZLLentz: I've added an issue for this, I might get around to it at some point: https://github.com/pcdshub/engineering_tools/issues/147

ZLLentz (approved):
I think this looks good/clean/useful :+1:

ZLLentz:
@klauer anything else to change here before merging?

wwright-slac:
Thanks @ZLLentz and @klauer !

PR #148, opened 2023-03-14
Added extra bracket to fix issue with last elif statement

wwright-slac:
## Description
The last elif statement was executing oppositely to what was intended. Replacing [ with the [[ operator fixed the issue.

### DIFF
--- a/scripts/new_bare_repo
+++ b/scripts/new_bare_repo
@@ -31,13 +31,13 @@ REPO=$2
 CWD=$PWD
 
 set -e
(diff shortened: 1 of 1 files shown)


klauer (approved):
Now that you point it out, that does look better - oops. I'd recommend in the future you give `shellcheck` a try for linting your scripts.

I've written a fair number of bash scripts and make very simple mistakes more often than I'd like to admit - shellcheck tends to reduce debugging time of those mistakes rather drastically.

PR #152, opened 2023-05-12
add spyder bash script and a line for finding duplicates was added to…

patoppermann:
… the epicsArchChecker script.

## Description
In order to be able to use the Spyder IDE, a script has been added to enginerring_tools, which calls the correct environment. And for easier finding of duplicates in the epicsArchChecker script, the read in files are now displayed.

## Motivation and Context
We had a request for the use of IDEs in our Python environment.
Finding of duplicates in the epicsArchChecker script.

## How Has This Been Tested?
Spyder Skript was testet to start in the RIX enviroment and XCS.
EpicsArchChecker was tested locally with a sample file.

Documentation is in the code.

-->

### DIFF
--- a/scripts/epicsArchChecker
+++ b/scripts/epicsArchChecker
@@ -98,7 +98,8 @@ def read_file(filename):
     except FileNotFoundError as errorDetail:
         print(filename, ' File not found!')
         print(errorDetail)
-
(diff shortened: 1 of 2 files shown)


ZLLentz (approved):
I have no context on the epics arch checker but the spyder script should work

PR #154, opened 2023-06-02
DEV: initial commit for scripts to make gifs from rix camera feed

KaushikMalapati:
## Description

Added two files: h5_to_gif.py, which given a path to images in an hdf5 file stored in the format that h5_img_collect from rixpython will convert to a gif and store in cwd, and rix_image_saver, which given a number 1-7 will use h5_img_collect to save 25 frames from a crix-gige-0X camera at /reg/d/iocData/ioc-crix-gige-0X/hdf5/ .
## Motivation and Context

Not related to an open issue, but will help capture events on crix1-3 and ceate gifs to be used in presentations
## How Has This Been Tested?

I have tested normal use cases, but welcome additional testing and feedback.

N/A

### DIFF
--- a/scripts/makepeds
+++ b/scripts/makepeds
@@ -12,11 +12,9 @@ Make a pedestal file for offline use
 
 OPTIONS:
         -u|--user        
-               user (needs to be able to log into the psananeh/feh)
+               user (needs to be able to log into the s3df)
         -p|--post <text>  
                add <text> to elog post
-        -F|--ffb
-                run job on FFB system
         -t|--test
                 do not deploy pedestals (epix10k only)
         -r|--run         
@@ -59,6 +57,8 @@ OPTIONS:
                 give path for alternative calibdir
         -b|--nbunch
                 number of bunches for the laseroff XTCAV calculations
+        --sdfdir
+                Directory where engineering tool in on S3DF. For development work only.
 EOF
 }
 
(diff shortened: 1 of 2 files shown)


scripts/h5_to_gif.py
```diff
@@ -0,0 +1,27 @@
...
+    return
+
+
+if __name__ == "__main__":
+    # Arguments
+    parser = argparse.ArgumentParser()
+    parser.add_argument('--file', '-f', help='hdf5 file containing images',
+                        required=True)
+    parser.add_argument('--speed', '-s', type=int, nargs='?', default=120,
+                        help='playback speed for animated gifs, default 120')
```
ZLLentz, line 25: Is there a unit for this? How fast is 120?

scripts/rix-image-saver
```diff
@@ -0,0 +1,23 @@
...
+HERE='/reg/g/pcds/pyps/apps/hutch-python/rix/'
+source "${HERE}/rixenv" > /dev/null
+# Hutch python config file
+cfg=$(mktemp)
+echo -e "hutch: rix\nload: rix.beamline" >> $cfg 
+# Script to save images
+script=$(mktemp)
+echo -e "h5_img_collect('crix0$cam', 25, path=\"/reg/d/iocData/ioc-crix-gige-0$cam/hdf5/\", file_name=\"crix0$cam-$time\")" >> $script
+# Launching non-interactive session with above config and script
+hutch-python --cfg $cfg $script > /dev/null
```
ZLLentz, line 20: Maybe the h5_img_collect script needs to be brought into engineering_tools as well. Loading the entire RIX environment to collect images is pretty heavy, and maybe other hutches would benefit from this script too.

This doesn't need to be your action item, but I think it should happen.

scripts/rix-image-saver
ZLLentz, line 1: Does a script that only works in rix belong in engineering_tools? @jyotiphy @silkenelson

scripts/rix-image-saver
```diff
@@ -0,0 +1,23 @@
...
+fi
+# Setting up environment
+HERE='/reg/g/pcds/pyps/apps/hutch-python/rix/'
+source "${HERE}/rixenv" > /dev/null
+# Hutch python config file
+cfg=$(mktemp)
+echo -e "hutch: rix\nload: rix.beamline" >> $cfg 
+# Script to save images
+script=$(mktemp)
+echo -e "h5_img_collect('crix0$cam', 25, path=\"/reg/d/iocData/ioc-crix-gige-0$cam/hdf5/\", file_name=\"crix0$cam-$time\")" >> $script
```
ZLLentz, line 18: I'd like to note that this script only works for cams with a very narrow name pattern, but I don't see an obvious reason for why this should be the case.

ZLLentz (commented):
Hi, I've given this a review. I have a lot of comments/food for thought. I don't think any of these are blockers though, the script works after all. The main thing to fix is documentation: the scripts need to be added to the readme and there are some parts I've noted that should be documented better for future usage.

scripts/h5_to_gif.py
```diff
@@ -0,0 +1,27 @@
+import argparse
+import h5py
+from PIL import Image
+
+def convert(file, speed=120):
+    """Convert images in an HDF5 file into a gif and save to cwd"""
+    # Getting images from HDF5 file
+    h5_file = h5py.File(file, 'r')
+    images = h5_file['entry']['data']['data']
```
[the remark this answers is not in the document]
KaushikMalapati: Yes, it assumes that the file came from h5_image_collect (or a file with images located at the leaf of the hdf5 file tree). They can different structures, and I noted this in the docstring.

scripts/h5_to_gif.py
```diff
@@ -0,0 +1,27 @@
...
+    # Converting to PIL.Image
+    images = [Image.fromarray(i).convert() for i in images]
+    # Saving as GIF
+    images[0].save(file.split('/')[-1].replace('.h5', '.gif'),
```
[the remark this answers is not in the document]
KaushikMalapati: I have changed this to append the .gif extension instead of replacing a .h5 extension (if present).

scripts/h5_to_gif.py
KaushikMalapati: I have changed the name to duration and noted that the number corresponds to how long each frame of the gif lasts in milliseconds.

KaushikMalapati:
I have replaced rix-image-saver with image-saver.py, with the same functionality but in python. I am using a temporarily generated config file to load a subset of the hutches environment that includes its cameras and the h5_img_collect function. I believe it will work on any hutch that has h5_img_collect in its environment. I also added an option to take images with buttons to capture anything time-sensitive.

scripts/image-saver.py
```diff
@@ -0,0 +1,143 @@
...
+        exit(f"Must 'source /reg/g/pcds/pyps/apps/hutch-python/{args.hutch}"
+             f"/{args.hutch}env' before running")
+    # Create hutch-specific config via temp file
+    config = NamedTemporaryFile()
+    config.write(f'hutch: {args.hutch}\nload: {args.hutch}.beamline'.encode())
+    config.seek(0)
+    # Load devices and hutch_python package
+    from hutch_python.load_conf import load
+    try:
+        objs = load(cfg=config.name)
```
ZLLentz, line 132: A future version of this could be even quicker by leveraging the internals in https://github.com/pcdshub/hutch-python/blob/master/hutch_python/cam_load.py to load only the camera and not the rest of the beamline too, but it's not possible right now because `h5_img_collect` is living in `rix3` instead of in engineering tools, so the beamline basically needs to be loaded anyway.

ZLLentz (approved):
Thank you for the detailed docstrings and readme additions! The code in general makes a lot more sense to me now too, it's much more direct and feature-rich.

Somebody else who plans on using the gui for their workflow should comment/review the screen, but the code looks reasonable. Maybe I could nitpick a bit but it seems unproductive to.

The biggest thing that could help these scripts is including h5_img_collect standalone in engineering_tools so we could make the startup more succint.

scripts/image-saver.py
KaushikMalapati: I have updated image-saver; it uses cam_load to load an individual camera, and I added a function based on CamH5.collect to collect/save images without using h5_img_collect. Is significantly faster and less hutch-dependent, since I do not need to source the hutch-python environment for a specific hutch or load every device or every camera, only the one specified as an argument. Using it without the -g flag has pretty much the same functionality as h5_img_collect. I am using /reg/g/pcds/pyps/apps/hutch-python/{hutch}/dev/hutch-python/hutch_python/cam_load.py and /reg/g/pcds/pyps/config/{hutch}/camviewer.cfg but these are in most if not all hutches.

scripts/image-saver.py
```diff
@@ -1,85 +1,179 @@
...
+
+    Returns
+    -------
+    camera: pcdsdevices.areadetector.detectors.PCDSAreaDetector
+        Camera object created by build_cam with hdf51 plugin
+        member to take and save images.
+    """
+    # Get cam_load functions
+    syspath.append(f'/reg/g/pcds/pyps/apps/hutch-python/{hutch}'
+                   '/dev/hutch-python/hutch_python/cam_load.py')
```
ZLLentz, line 95: This directory is not guaranteed to exist, it should be expected that the dev folder will be periodically cleared as fixes propagate to tagged libraries.

If the script is ran using the normal central conda environments (`source /cds/group/pcds/pyps/conda/pcds_conda`) or from the hutch python envs (`source /cds/group/pcds/pyps/apps/hutch-python/{hutch}/{hutchenv}`) then `hutch_python.cam_load` should be importable without any path appending.

`/reg/g` and `/cds/group` are network mounts to the same disk, where `/reg/g` is the old/deprecated path.

scripts/image-saver.py
```diff
@@ -1,85 +1,179 @@
...
+    from hutch_python.cam_load import interpret_lines, build_cam
+    # Get config for cam_name
+    try:
+        f = open(f'/reg/g/pcds/pyps/config/{hutch}/camviewer.cfg')
```
ZLLentz, line 99: It's considered standard practice in python to always call `open` inside a context manager to ensure the file handle is closed in a timely manner, for example:

```
with open(filename, 'r') as fd:
    lines = fd.readlines()
```

This causes the file handle to be closed immediately after this line, instead of at an indeterminate time later when the python gc collector decides to clean up the file descriptor object (in your code, this is `f`)

scripts/image-saver.py
```diff
@@ -1,85 +1,179 @@
...
+    except Exception as e:
+        print(e)
+        exit('Could not open camera config file, '
+             f'/reg/g/pcds/pyps/config/{hutch}/camviewer.cfg')
+    cam_line = [line for line in f.readlines() if 'im2k0'.upper() in line]
```
ZLLentz, line 104: I think this line must be in error, there are plenty of cameras that aren't named `im2k0`

scripts/image-saver.py
```diff
@@ -1,85 +1,179 @@
...
+        nonlocal collecting
+        collecting = value
+
+    def _writing_cb(value, **kwargs):
+        """Get a callback when writing changes"""
+        nonlocal writing
+        writing = value
+    # Start collecting and saving
+    collect_uid = camera.hdf51.capture.subscribe(_collecting_cb)
+    write_uid = camera.hdf51.write_file.subscribe(_writing_cb)
```
ZLLentz, line 59: It's unlikely, but the code here could hit a race condition if the camera IOC doesn't behave as you'd expect- for example, if these callbacks get run based on the values from a previous function call.

ZLLentz (commented):
Just a few small things I noticed. I very much appreciate how this update divorces the script from the hutch-python loading process, which can take a long time to set up resources for the interactive session.

I'm surprised this PR is still up- we probably could have merged it prior to your last change and reviewed the newer update separately.

Again, someone else should give feedback on the actual GUI, preferably someone who plans to run this as part of their workflows.

scripts/image-saver.py
KaushikMalapati: Good to know, can I assume that this will be run in either of those environments or should I check for/enforce that somehow?

ZLLentz: I think there are two sensible approaches:
- "This is a python script that requires `pcds_conda`, please source `pcds_conda` before running this."
- We can write a shell script that sources pcds_conda and then runs this python script

scripts/image-saver.py
KaushikMalapati: Not sure I understand how a race condition could be caused by a previous function call, since each time the script is run a new camera object is made with its own hdf51 member, and since I am not multithreading there will only ever be one of each of the two subscribers. I (briefly) looked at some of the implementation of EpicsSignalWithRBV in ophyd/signal.py, and I think that if you had multiple subscribers, (in different processes) there could be undefined behavior, but I don't see a clean solution since you would need some way of obtaining pv locks at the channel access level. Almost everything hdf51-related in take_images is from CamH5.collect, which is what h5_img_collect uses, which is why I didn't suspect any issues. I will look into this more, but if you have any more thoughts on what I could do to try and fix this, I'm all ears.

ZLLentz: I'm not sure if you absolutely need to fix this, since it will be very rare and I can't prove that it will happen. This is the sort of thing where I'm more bringing awareness that "code that looks like this can be suspect to race conditions".
Your script and all the objects inside it are allocated each time it is run, but the camera and its EPICS IOC continue to exist and hold some state information.

Consider the following:
1. You run this script a few times
2. On one of the runs, the IOC lags a bit or the network has a bad time, and the IOC does not reset the `capture` and `write_file` PV values (or, the script doesn't see the reset) prior to these subscriptions being set up, so they are still in the final states from the previous script run.
3. The subscriptions see the stale values and conclude that the collections is already done, when it is not.

Again, this is just for awareness that these sorts of patterns can cause race conditions for scripts that interact with EPICS IOCs.

KaushikMalapati: Thanks for the clarification. If I wanted to fix this, or similar situations that would occur when interacting with EPICS IOCs, how would I go about doing that? Would using Archiver data or polling for the hdf5 file's creation be a better way to do this? Also should I be worried about multiple processes simultaenously using the hdf5 plugin for the same camera and interfering with each other?

ZLLentz: For similar situations I would probably do one of or some combination of the following depending on the situation:
- Start the subscription prior to making any changes so it always starts in the same state
- use the `old_value` callback kwarg in addition to the `value` callback kwarg
- rely on the `timestamp` callback kwarg

I think for this specific situation polling the hdf5 file's creation is a pretty reasonable solution too, that shouldn't ever happen out of sequence (but I guess it could take some moments to propagate through the network file system?)

I think your worry about multiple processes is valid. I'm not entirely sure how to approach that one. In general we'd want to make some multi-process mutex but in this case you're not actually in control of the other processes and you're not guaranteed that this is the only script that will use the hdf5 plugin... I think it's OK to disregard it in favor of putting your efforts elsewhere, but I'd approach it from the angle of trying to check if someone else is using the plugin and giving a reasonable error message.

ZLLentz (approved):
I think this is good to go!

Note I haven't run the latest code myself- there's no obvious way for me do to this without interfering in ongoing beamline operations. So I'm trusting that you've tested it. The code is plenty clean.

If you haven't already gotten feedback from beamline users about the GUI I suggest you do that too, usually people have strong opinions about GUIs!

Let if me know if you plan on doing anything else or if I should click merge.

KaushikMalapati:
Mostly gui changes this time, some small tweaks in other places. No rush to review. If at all looks good, feel free to merge.

ZLLentz:
Hold up: it looks like a bunch of `makepeds` commits from Vincent that were in master got included into this PR, but not as a fast forward, instead as a bunch of duplicate commits with different hashes. What happened?

KaushikMalapati:
I have opened a new pull request (#161 ). Could you add yourself as a reviewer to it? I don't seem to have permission to do so.

ZLLentz:
I've reviewed and merged the other PR :+1:

PR #156, opened 2023-06-19
Pedestals on S3DF

vespos:
## Description
Make makepeds compatible with running on the S3DF, without changing the well established workflow

## How Has This Been Tested?
Tested locally S3DF for data on the FFB.
Need more test for data on S3DF, from hutch opr accounts and with data in transfer

Docstrings in scripts

### DIFF
--- a/scripts/makepeds
+++ b/scripts/makepeds
@@ -200,11 +200,6 @@ else
     HUTCH=${EXP:0:3}
 fi
 
-if [[ $HOSTNAME =~ "sdf" ]] && [[ $DIR =~ "/reg/g/pcds" ]]; then
-    echo DEBUG change engineering_tools directory if standard hutch one
-    #DIR=`echo $DIR | sed s/\reg\g\pcds/\cds\group\lcls\ds/tools/g`
-fi
-    
 if [[ $HOSTNAME =~ "psana" ]] || [[ $HOSTNAME =~ "drp-srcf" ]] || [[ $HOSTNAME =~ "sdf" ]]; then
     #echo $DIR/makepeds_psana $@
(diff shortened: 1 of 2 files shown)


scripts/makepeds
```diff
@@ -222,7 +222,7 @@ else
     elif [[ $FFB == 2 ]]; then
         echo calling on SDF system: makepeds_psana $@
 	#we need to change paths here as directories are completely independent
-	DIR=`echo $DIR | sed 's/\/cds\/group\/pcds/sdf\/group\/lcls\/tools/g' | sed 's/\/reg\/g\/pcds/\/sdf\/group\/lcls\/ds\/tools/g'`
+	DIR=`echo $DIR | sed 's/\/cds\/group\/pcds/\/sdf\/group\/lcls\/tools/g' | sed 's/\/reg\/g\/pcds/\/sdf\/group\/lcls\/ds\/tools/g'`
```
silkenelson, line 223: Does the 'cds' replace not also need the 'ds' at lcls/ds/tools?

silkenelson: Also, lines 205-208 can go as they are doing what lines 2245 et al now do better. I could not figure out how to comment there.

silkenelson (commented):
Minor comments.

scripts/makepeds
vespos: That's right!
ANd I will check the lines you mention. You mean line 245 though, right?

silkenelson (approved):
I think we only need 1 cpu-per-task, but I had not finished testing. Either way, given the total number of cores, they should be able to fit into a single node when run non-exclusively, so this is a step forward.

PR #158, opened 2023-06-23
Adding default camera for UED

antglr:
## Description
Added a condition to the elseif for UED

## Motivation and Context
Because CamViewer was not running for UED  (https://jira.slac.stanford.edu/browse/LCLSECSD-2108?filter=-2)

## How Has This Been Tested?
Running the command

Not needed

### DIFF
--- a/scripts/camViewer
+++ b/scripts/camViewer
@@ -292,7 +292,6 @@ else
 fi
 
 
-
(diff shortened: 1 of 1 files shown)


silkenelson (approved):
Simple enough.

PR #159, opened 2023-07-12
clean up psana / ffb stuff from makepeds

vespos:
## Motivation and Context
As we are now using the S3DF, takepeds and makepeds should only run from there, or the calib files won't be written to the right place. Removing old psana and ffb from it ensures that there won't be accidental run of makepeds on the wrong systems. Should older pedestal be reprocessed, one should run from a previous release of engineering_tools

## How Has This Been Tested?
Full workflow tested with the Jungfrau from xpp-control

Self documented with docstring

### DIFF
--- a/scripts/makepeds
+++ b/scripts/makepeds
@@ -193,7 +193,8 @@ else
     if [[ $S3DF == 1 ]]; then
         echo calling on SDF system: makepeds_psana $@
         #we need to change paths here as directories are completely independent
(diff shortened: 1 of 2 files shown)


patoppermann (commented):
psana was changed to s3df, looks good.
Also, milano Nodes are added.

patoppermann (approved):
All good to go.

PR #160, opened 2023-07-17
default to s3df even if no queue is passed

vespos:
Makepeds to work on S3DF in interactive mode.

## How Has This Been Tested?
With xcsx1004322 run 16

N/A

### DIFF
--- a/scripts/takepeds
+++ b/scripts/takepeds
@@ -32,7 +32,3 @@ echo $PYCMD -i "${HUTCH^^}" -u `whoami` -e "$EXP"  -t DARK  -r $RUN -m "$elogMes
 $PYCMD -i "${HUTCH^^}" -u `whoami` -p pcds -e "$EXP"  -t DARK  -r $RUN -m "$elogMessage"&
 
 echo 'Please call: makepeds -q milano -r '`get_lastRun`' -u <userID>'
-echo 'To run job in the interactive node run (slower): makepeds -r '`get_lastRun`' -u <userID>'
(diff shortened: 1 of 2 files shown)


scripts/takepeds
```diff
@@ -31,8 +31,8 @@ HUTCH=${EXP:0:3}
 echo $PYCMD -i "${HUTCH^^}" -u `whoami` -e "$EXP"  -t DARK  -r $RUN -m "$elogMessage"
 $PYCMD -i "${HUTCH^^}" -u `whoami` -p pcds -e "$EXP"  -t DARK  -r $RUN -m "$elogMessage"&
 
-echo 'please call: makepeds -r '`get_lastRun`' -u <userID>'
-echo 'for gain-switching detectors we recommend: makepeds -u <userID> -q milano -r '`get_lastRun`
+echo 'Please call: makepeds -q milano -r '`get_lastRun`' -u <userID>'
+echo 'To run job in the interactive node run (slower): makepeds -r '`get_lastRun`' -u <userID>'
```
silkenelson, line 35: Actually, for non-gainswitching detectors it's really not clear to me if interactive is faster or slower than the queue as this'll depend on the other uses of the node where the job runs & the SLURM overhead. Maybe just don't say anything and assume that people know that not passing a queue will make the job run interactively (which we'll write on the pedestal page)?

silkenelson (approved):
I'm ok with the new takepeds message, but I think I may prefer just leaving the second line out.

scripts/takepeds
vespos: I agree, let's take that first option out altogether!

silkenelson (approved):

PR #161, opened 2023-07-26
Adding scripts to take camera images and convert saved hdf5 files to gifs

KaushikMalapati:
I don't know what happened on the original pull request. I believe I accidentally pulled on my rix-gifs branch instead of master or rebased improperly and duplicated Vincent's commits. I wanted to revert but ended up closing the pull request in the process. Going to try again with a new request. hdf5_to_gif.py has not changed. I changed the readme entry for image-saver to reflect the gui changes I made. The changes in image-saver are mainly in button_gui, based on feedback from future users. They are turning the button green while images are being taken, being able to change the number of images to take from within the gui, and being able to change the text on each button (they start as just the name on the camera). There are also minor changes to main and take_images which involve filenames; now the filename argument will only be a prefix, and the camera name and timestamp will always be appended, which will prevent overwrites incase of running the script without -g/--gui multiple times with the same camera and filenames. This has made the take_images wrapper obsolete. Finally, multiple cameras can now be provided to the -c/--cam argument; if in gui mode, this will result in mutliple buttons, otherwise will take images with every camera specified (filenames will still be unique since camera name will be appended to each). Sorry for the trouble, I'll be more careful when pulling/pushing in the future. Most of my changes are fairly straightforward, and flake8 has no complaints. Let me know what else I should od whenever you have time to look over this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -270,7 +270,8 @@ usage: grep_ioc KEYWORD [hutch]<br/>
      Command line arguments -c, -n, -p, and -f to specify camera name, number<br/>
      of images, path to save hdf5 file to, and name to save hdf5 file as.<br/>
      Also can use -g switch to open a GUI with a button that when pressed takes<br/>
-     images with specified parameters - can be pressed multiple times.<br/>
+     images with specified parameters - can be pressed multiple times. The number<br/>
+     of images and the label on each button can be changed within the gui.<br/>
     </td>
 </tr>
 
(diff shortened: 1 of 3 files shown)


ZLLentz:
It's OK, thank you for taking the time to clean up the git stuff!
I'm going to approve and merge, the implementation is still :+1:

ZLLentz (approved):

PR #162, opened 2023-08-23
wherepsana typo: tmp should be /tmp

mcb64:
wherepsana creates a temporary file... this should be in /tmp, not in the current directory tmp.


silkenelson (approved):
lgtm.

PR #163, opened 2023-09-08
feat(set_gem_timing): allowing operators to switch between HXR-GEM SC and NC timing with a one line script

joshc-slac:
Adding what should be a temporary hack script but may be somewhat long lived to provide a one line mechanism for operators to switch the timing paradigm on the HXR-GEM. This script:
- Sets PV: `KFE:CAM:TPR:02:MODE` to either SC or NC
- Invokes `imgr` to set the GEMs IOC version to the correct version for the corresponding timing paradigm

Invokation once sourced is of form: `set_gem_timing [scheme]` 
Schemes besides `NC` and `SC` will be rejected.

Associated PR: https://github.com/pcdshub/IocManager/pull/9

## Description
The `set_gem_timing` bash script critically:
- Sets PV: `KFE:CAM:TPR:02:MODE` to either SC or NC
- Invokes `imgr` to set the GEMs IOC version to the correct version for the corresponding timing paradigm

### Argument and State Parsing
- Input validates that only 1 argument is supplied and that argument is either `SC` or `NC`
- Will fail if `KFE:CAM:TPR:02:MODE` is already in whatever mode is trying to be actuated
- Will fail if `imgr --info` determines that the requested software is already on the device

## Motivation and Context
Ticket [ECS-3765](https://jira.slac.stanford.edu/browse/ECS-3765) requests a script to switch timing of the HXR-GEMs for the NC and SC timing schemas.

For the SC timing schema the crate that contains the HXR-GEM cannot be upgraded (or is prohibitively expensive to upgrade) which means the non SLAC EVR cannot receive / decode LCLS2 timing pulses.

To circumvent this Silke hardwired the LCLS2 timing signal from the IM2K0 to the HXR-GEM enabling LCLS2 timing to be passed to the L, hard x-ray, side of things. This means that as long as the hack exists and we want to collect SC (LCLSII) data on the HXR-GEM we need to toggle the IOC software to be the one that ignores timing and triggers when the IM2K0 camera triggers and we need a mechanism to toggle it back when we want to send NC beam to the HXR-GEM and get the additional information (timestamps) provided by actually going through the EVR on the crate.

This toggle can be accomplished [manually](https://confluence.slac.stanford.edu/pages/viewpage.action?spaceKey=PCDS&title=LCLS+II+Timing+Switchover+for+GEMs):

- From machine in the CA network: caput KFE:CAM:TPR:02:MODE X where X == NC or SC
  - I have heard but not confirmed this sets a bunch of PVs to predetermined values
- Switch the running software via imgr
  - ioc/fee/GasDetDAQ/R4.0.22 <- for NC
  - ioc/fee/GasDetDAQ/R4.0.22-NOTS <- SC

This script aims to deprecate both manual mechanisms GUI and CLI in switching between timing paradigms in favor of a one line command. This command can be plumbed into a button if desired.

## How Has This Been Tested?
- Josh has used it to switch timing paradigms for ACR for the evening shift on 9/6 and both shifts on 9/7.

## TODO:
- [x] this has all been from Josh's user account 
  - [x] check permissioning works as intended on operator account
  - [x] determine if sourcing should be handled by operator or this script
- [x] Upgrade the imgr instance for the lfe hutch to contain changes made in https://github.com/pcdshub/IocManager/pull/9
- [x] I would like to remove the absolute filepaths; but I would like to do it in a way thats principic to the rest of our system, someone will likely have to show me how we do it here

* https://confluence.slac.stanford.edu/display/PCDS/IocManager+R2-5-0

### DIFF
--- a/scripts/set_gem_timing
+++ b/scripts/set_gem_timing
@@ -9,7 +9,7 @@ IOC_PATH=0
 
 # User input validation
 if [[ -n $1 ]]; then # if the user passed at least one arg
-  if [[ $1 != $NC_STRING ]] && [[ $1 != $SC_STRING ]]; then #validate argument is either SC or NC
+  if [[ $1 != "$NC_STRING" ]] && [[ $1 != "$SC_STRING" ]]; then #validate argument is either SC or NC
     echo "Error. Argument must be either: [NC or SC]"
     exit 1
   else
@@ -26,7 +26,7 @@ if [[ -n $2 ]]; then
 fi
 
 #get caget and caput
-source /cds/group/pcds/setup/epicsenv-cur.sh #Note: (josh) we should only source the actual executables we are using, this is heavy handed. Should be considered suboptimal, but should keep us consitent with the rest of the stack
+source /cds/group/pcds/setup/epicsenv-cur.sh #Note: (josh) we should only source the actual executables we are using, this is heavy handed. shellcheck does not love this; however if this path doesnt exist we have bigger problems
 #get imgr
 IMGR_CMD_PATH="/reg/g/pcds/engineering_tools/latest-released/scripts/imgr"
 
@@ -35,7 +35,7 @@ OP_MODE=$(caget KFE:CAM:TPR:02:MODE | awk '{print $2}')
 SET_MODE_CMD="caput KFE:CAM:TPR:02:MODE $1"
 
 echo "Current OP_MODE is: $OP_MODE"
-if [[ $OP_MODE = $1 ]]; then
+if [[ $OP_MODE = "$1" ]]; then
   echo "Already in requested mode: $OP_MODE exitting with NOP"
   exit 1
 else
@@ -44,7 +44,7 @@ else
 fi
 
 # only now discern which mode we want to put it in, TODO(josh): theres some more bashy way of string unwrapping todo this....
-if [[ $1 = $NC_STRING ]]; then
+if [[ $1 = "$NC_STRING" ]]; then
   IOC_PATH=$IOC_PATH_NC
 else
   IOC_PATH=$IOC_PATH_SC
@@ -52,7 +52,7 @@ fi
 
 SET_IOC_PATH_CMD="${IMGR_CMD_PATH} ioc-lfe-gasdet-daq --upgrade $IOC_PATH"
 CURR_IOC_PATH=$(${IMGR_CMD_PATH} ioc-lfe-gasdet-daq --info | awk 'FNR == 4 {print $3}') #Warning: make sure your release of imgr supports --info 
-if [[ CURR_IOC_PATH = IOC_PATH ]]; then
+if [[ $CURR_IOC_PATH = "$IOC_PATH" ]]; then
   echo "Current IOC path is: $CURR_IOC_PATH which is equiavlent to requested IOC path for the requested switch to $1, not invoking imgr"
 else
   echo "Actuating ioc-lfe-gasdet-daq IOC software change from $CURR_IOC_PATH to: $IOC_PATH"
--- a/scripts/set_gem_timing
+++ b/scripts/set_gem_timing
@@ -1,0 +1,57 @@
+#!/bin/bash
+
+NC_STRING="NC"
+SC_STRING="SC"
+# TODO:(josh) reflect on hardcoded releases
+IOC_PATH_NC="ioc/fee/GasDetDAQ/R4.0.22"
+IOC_PATH_SC="ioc/fee/GasDetDAQ/R4.0.22-NOTS"
+IOC_PATH=0
+
+# User input validation
+if [[ -n $1 ]]; then # if the user passed at least one arg
+  if [[ $1 != $NC_STRING ]] && [[ $1 != $SC_STRING ]]; then #validate argument is either SC or NC
+    echo "Error. Argument must be either: [NC or SC]"
+    exit 1
+  else
+    echo "User passed argument of: $1"
+  fi
+else
+  echo "Error. Must pass at least one argument"
+  exit 1
+fi
+
+if [[ -n $2 ]]; then
+  echo "Warning. Must pass strictly one argument, additional arg: $2 is ignored"
+  exit 1
+fi
+
+# check current operational state
+OP_MODE=$(caget KFE:CAM:TPR:02:MODE | awk '{print $2}')
+SET_MODE_CMD="caput KFE:CAM:TPR:02:MODE $1"
+
+echo "Current OP_MODE is: $OP_MODE"
+if [[ $OP_MODE = $1 ]]; then
+  echo "Already in requested mode: $OP_MODE exitting with NOP"
+  exit 1
+else
+  echo "Actuating requested timing change, caputing $1 to KFE:CAM:TPR:02:MODE"
+  $SET_MODE_CMD
+fi
+
+# only now discern which mode we want to put it in, TODO(josh): theres some more bashy way of string unwrapping todo this....
+if [[ $1 = $NC_STRING ]]; then
+  IOC_PATH=$IOC_PATH_NC
+else
+  IOC_PATH=$IOC_PATH_SC
+fi
+
+SET_IOC_PATH_CMD="imgr ioc-lfe-gasdet-daq --upgrade $IOC_PATH"
+CURR_IOC_PATH=$(imgr ioc-lfe-gasdet-daq --info | awk 'FNR == 4 {print $3}') #Warning: make sure your release of imgr supports --info 
+if [[ CURR_IOC_PATH = IOC_PATH ]]; then
+  echo "Current IOC path is: $CURR_IOC_PATH which is equiavlent to requested IOC path for the requested switch to $1, not invoking imgr"
+else
+  echo "Actuating ioc-lfe-gasdet-daq IOC software change from $CURR_IOC_PATH to: $IOC_PATH"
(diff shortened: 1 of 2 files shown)


scripts/set_gem_timing
```diff
@@ -0,0 +1,57 @@
+#!/bin/bash
+
+NC_STRING="NC"
+SC_STRING="SC"
+# TODO:(josh) reflect on absolute paths
+IOC_PATH_NC="/reg/g/pcds/epics/ioc/fee/GasDetDAQ/R4.0.22"
+IOC_PATH_SC="/reg/g/pcds/epics/ioc/fee/GasDetDAQ/R4.0.22-NOTS"
```
ZLLentz, line 7: I think canonically in bash scripts you'd use environment variables to generalize something like this (maybe someone has a better suggestion?), but I don't think you need to worry about it here. Given the expected lifetime of the mode of operations, the only risk I see is if we need to update the ioc versions and then we'd need to come back and update this script again (and also make sure people stop using the old version of the script...)

scripts/set_gem_timing
```diff
@@ -0,0 +1,57 @@
...
+IOC_PATH=0
```
ZLLentz, line 8: What's the purpose of setting the variable here? It isn't used until it is overridden later on.

scripts/set_gem_timing
```diff
@@ -0,0 +1,57 @@
...
+fi
+
+if [[ -n $2 ]]; then
+  echo "Warning. Must pass strictly one argument, additional arg: $2 is ignored"
+  exit 1
+fi
+
+# check current operational state
+OP_MODE=$(caget KFE:CAM:TPR:02:MODE | awk '{print $2}')
+SET_MODE_CMD="caput KFE:CAM:TPR:02:MODE $1"
```
ZLLentz, line 30: What's the purpose of breaking this out as a string? I don't see anywhere else where this is used.

scripts/set_gem_timing
```diff
@@ -0,0 +1,57 @@
...
+
+echo "Current OP_MODE is: $OP_MODE"
+if [[ $OP_MODE = $1 ]]; then
+  echo "Already in requested mode: $OP_MODE exitting with NOP"
```
ZLLentz, line 34: Is it possible that the user might have gotten into an inbetween state where the PVs have been set to the requested mode, but the IOC has not been switched over yet? If so, I think we should proceed regardless instead of exiting here.

scripts/set_gem_timing
```diff
@@ -0,0 +1,57 @@
...
+fi
+
+# only now discern which mode we want to put it in, TODO(josh): theres some more bashy way of string unwrapping todo this....
+if [[ $1 = $NC_STRING ]]; then
+  IOC_PATH=$IOC_PATH_NC
+else
+  IOC_PATH=$IOC_PATH_SC
+fi
+
+SET_IOC_PATH_CMD="imgr ioc-lfe-gasdet-daq --upgrade $IOC_PATH"
```
ZLLentz, line 48: Same question as above: why is this broken out as a string prior to executing it?

scripts/set_gem_timing
ZLLentz, line 1: General bash tip: I recommend running `shellcheck` on your script to see if there's anything it can check for you automatically. This is a free web tool and it's also included in our `pcds_conda` python environment.

ZLLentz, line 1: Also: consider updating the repo readme to include some information about your new script.

ZLLentz (commented):
Just a few minor comments, suggestions, and questions

ZLLentz:
Also, an aside: thank you for being detailed and clear in the PR description here, it made it much easier to understand the intent of the code. I think the script is overall very good.

silkenelson (approved):
I'll approve for now if we would prefer for ACR to use a release version rather than a home-dir. That said, maybe following a better process and including the comments Zach made would be cleaner.

ZLLentz:
I would not be opposed to a merge if the intent is "get the script out now, working, with a tagged version". I only found one potential edge case, which technically shouldn't ever happen.

scripts/set_gem_timing
joshc-slac: this is cool!

scripts/set_gem_timing
joshc-slac: I'll respond to both here. It's been a good few years since I've written any bash but I generally (especially in a low to no production gate system like we have here) will set my intended command to a string variable so I can echo it in testing as opposed to running. When I was cleaning up I somewhat intentionally left this artifact, if it obfuscates the code I am more than happy to remove. What do you think?

scripts/set_gem_timing
joshc-slac: Like the strings you see mainly as a mechanism to communicate to other devs.... thats probably somewhat ill principled and a comment would reflect that better

scripts/set_gem_timing
joshc-slac: Great point, I think an analog of this comment is in my TODOs section. I would like to understand how we do it here I just didn't know how much of the path should be environment-variable-ized. 

Agree with lifetime point; hopefully the TODO is reminder enough for us to address. If you would prefer tracking outside of source code let me know.

scripts/set_gem_timing
ZLLentz: Gotcha, that makes sense. There's no need to remove it, I was simply curious.

scripts/set_gem_timing
ZLLentz: It's in the source code as a comment and there's a record of discussions here and in the PR text, so that should be enough.

scripts/set_gem_timing
silkenelson: We should probably start using the /cds/group instead of /reg/g
It'll be so hard to beat my muscle memory here...

scripts/set_gem_timing
joshc-slac: Even better news. @mcb64 let me know `imgr` infers release area, we do not need full path

joshc-slac:
Following up on the remaining TODO:

Based on this [slack feedback](https://slac.slack.com/archives/C02VBTF6QMT/p1694276928075339?thread_ts=1694225717.879579&cid=C02VBTF6QMT) it seems operators **do not** normally operate with dotfiles sourced. 

This makes sense, so I will make the script source the needed elements for them so it is truly a one stop shop. I have scheduled time with ACR to observe them using this tooling. Planning on closing this out after that confirmation.

joshc-slac:
> Following up on the remaining TODO:
> 
> Based on this [slack feedback](https://slac.slack.com/archives/C02VBTF6QMT/p1694276928075339?thread_ts=1694225717.879579&cid=C02VBTF6QMT) it seems operators **do not** normally operate with dotfiles sourced.
> 
> This makes sense, so I will make the script source the needed elements for them so it is truly a one stop shop. I have scheduled time with ACR to observe them using this tooling. Planning on closing this out after that confirmation.

Following up on this. Ran through utilizing this with ACR. Scripts seem sufficient for their needs.

There was a lingering complaint about ssh-tunneling. This could be resolved given sufficient GUI

silkenelson:
> > Following up on the remaining TODO:
> > Based on this [slack feedback](https://slac.slack.com/archives/C02VBTF6QMT/p1694276928075339?thread_ts=1694225717.879579&cid=C02VBTF6QMT) it seems operators **do not** normally operate with dotfiles sourced.
> > This makes sense, so I will make the script source the needed elements for them so it is truly a one stop shop. I have scheduled time with ACR to observe them using this tooling. Planning on closing this out after that confirmation.
> 
> Following up on this. Ran through utilizing this with ACR. Scripts seem sufficient for their needs.
> 
> There was a lingering complaint about ssh-tunneling. This could be resolved given sufficient GUI

I would skip that. If ACR wants it more convenient, they can make take are of it - I believe Matt Gibbs made them something easier for the PMPS GUI.

scripts/set_gem_timing
ZLLentz: I think this was my only important suggestion, and it's also the only one skipped over. I still maintain that we shouldn't immediately exit if the PV is already changed.

silkenelson: I thought I had looked at that and somehow did not see it still did that. I agree that we should not quit when one of the PVs is changed because the IOC may not be changed yet. The cleanest way would be to get the running version and only update if it's not what we want in the requested mode.

joshc-slac: Thanks for following up on this, sorry for missing it. I'll make the patch Silke recommended.

silkenelson: Just not exiting is the quicker solution, but it may cause additional IOC reboots which do take a while. But - safety by obscurity?

silkenelson: Unrelated comment: currently you exit with the second argument, while the text just tells users it'll be ignored. Super minor.

joshc-slac: I see what you're saying. When I wrote this I treated the PV as source of truth and you are saying it should instead by IOC software because that has a longer boot time. From my work with the operators the past week they seem very competent, I'd imagine like you allude to above gating may actually induce additional edge cases. 

They seem happy to use this tool as more of a hammer or hack to clobber the HXR GEM into the timing schema they want. 

Exiting with the second argument was done to be explicit about what is ignored, when invoking CLI arguments I have found people sometimes have difficulty counting (may be a relic of `char * argv[]` things as a joke to myself).

PR #165, opened 2023-09-18
Cater to restricted experiments

slacmshankar:
With run 21, we introduced the notion of restricted experiments; that is, experiments for which most LCLS staff do not have permission. We get a permission denied exception when getting run table data from these experiments. We should just try/catch the exception, log it as expected and move on.


ZLLentz:
Murali, what's going on with the source of this PR? Is this from a fork that's also in the pcdshub org? This is a confusing way to do this sort of thing.

PR #166, opened 2023-09-25
[ENH] Additional `makepeds` reporting and exclusion of recorder data stream

gadorlhiac:
## Description

- Currently, `makepeds` waits until jobs complete, checking every 10 seconds, but does not indicate the job status (pending/running/etc). This PR provides feedback on the status of those submitted jobs so a decision can be made as to whether it's just a matter of waiting patiently or requesting additional resources (etc.).
- Currently, most portions of `makepeds` exclude the recorder datastream; however, the `calibrun` portion does not. This PR excludes the recorder data stream from this latter part as well.
- `Ctrl-c` can now be used to cancel all SLURM jobs. This allows early exit if jobs are stuck pending and a new strategy is needed.
- Exits `makepeds` early if `sbatch` fails and provides a hint to a common cause (account doesn't have permission on experiment).

## Motivation and Context

- Provides additional information to make decisions about compute resources. Resolves [[ECS-3879]](https://jira.slac.stanford.edu/browse/ECS-3879?filter=-1)
- Helps to improve the efficiency of `makepeds` by only considering events with usable detector information. Resolves [[ECS-3770]](https://jira.slac.stanford.edu/browse/ECS-3770)
- Provide a way to easily cancel jobs when a new strategy is needed.
- `makepeds` reports a lot of information and failures of `sbatch` can sometimes be missed. Early exit with a message can help catch this problem faster.

## How Has This Been Tested?

- Tested on experiment `xcsl1004821` run `258`.

- NA

image
Captura desde 2023-09-26 14-02-22

### DIFF
--- a/scripts/makepeds_psana
+++ b/scripts/makepeds_psana
@@ -62,6 +62,20 @@ check_running_jobs()
     REMJOBIDS=()
     for JOBID in "${JOBIDS[@]}"; do
         if  squeue | grep -q "$JOBID"; then
+            JOBSTATUS=`squeue | grep "${JOBID}" | awk '{print $5}'`
+            if [[ $JOBSTATUS == "R" ]]; then
+                echo "${JOBID} is RUNNING"
+            elif [[ "${JOBSTATUS}" == "PD" ]]; then
+                echo "${JOBID} is PENDING"
+            elif [[ "${JOBSTATUS}" == "CG" ]]; then
+                echo "${JOBID} is COMPLETING"
+            elif [[ "${JOBSTATUS}" == "PR" ]]; then
+                echo "${JOBID} has been PREEMPTED!"
+            elif [[ "${JOBSTATUS}" == "F" ]]; then
+                echo "${JOBID} has FAILED!"
+            else
+                echo "${JOBID} is STOPPED or SUSPENDED"
+            fi
             REMJOBIDS+=( "$JOBID" )
             NJOBS=$((NJOBS+1))
         fi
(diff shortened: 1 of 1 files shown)


silkenelson (commented):
They look correct to me, but we are running this check every 10 sec (current setting). We have 3,5 or maybe even 20+ jobs running at the same time, if each job reports its status the screen will become somewhat hard to read.

vespos:
Does this warn the user if the jobs are not submitted at all? 
This can happen sometimes, for example when the job is launched with an account that does not have the right to run with the exp account. Usually the error message for that is further up and the scientists do not see it.

gadorlhiac:
@silkenelson This is a fair point. One simple option might be to have a "bool" or "counter" that checks if any jobs are not running - e.g. it loops through the jobs and then only prints once per loop `Jobs PENDING`, or if more detail is needed `X jobs are NOT running`. This avoids a lot of clutter - from printing all states, and also from printing for all jobs. I think I need to test though to see how bad the clutter is when it's actually run as is.

@vespos I don't think this warns the user. There is a check afterwards for failed jobs but do you think an early check it submission is necessary?

vespos:
Yes, I think it would be good to have a check that jobs are actually submitted. 
The error message looks like this for the case I mention above: 
`sbatch: error: Batch job submission failed: Invalid account or account/partition combination specified`

I think we could detect the `sbatch: error` part, which I expect should be generic

silkenelson:
I'd assume in that case that the sbatch command itself would return an error code?
This should be before we call the 'check on the jobs' section.
Once jobs are running, what amount of detail should we get into? I'm not which possible messages beyond RUN and PENDING we'll actually see. I feel 'a' number would be good, but go back&firth between bringing how many jobs are running versus how many are still pending.

gadorlhiac:
@vespos Okay, this makes sense - I can add some error checks for the submission itself (maybe with a helpful hint like "Does this account have permission to run...")

@silkenelson For the latter 'check jobs' section maybe a two part message?
```
Running N jobs ...
[...]
No. Jobs RUNNING: X
No. Jobs PENDING/NOT RUNNING: Y
Checking again in 10 s
```

If we went with one single number I think I'd lean towards the PENDING number since if something goes wrong and things get really stuck, this number won't change very much so it will be easier to track and see that. What do you think?

silkenelson:
@vespos any preference - two numbers v one single number (for pending jobs)?

silkenelson (approved):
I'm basically ok with the request as it stands - there might be many lines for MEC, but let's fix it should be become a problem and for now we have more detailed feedback.

PR #167, opened 2023-09-26
ENH Updating image-saver gui based on user feedback to add/change cam…

KaushikMalapati:
…era objects and automake gifs

## Description

hdf5_to_gf.py now takes an optional argument specifying where to save the gifs to, defaulting to the cwd. image-saver.py, renamed to image_saver, has the ability to change which camera a button is connected to (previously this only changed the text displayed on the button) and the ability to add more buttons with more cameras. Additionally, it uses the convert method from hdf5_to_gif.py to automatically convert hdf5 images to gifs and place them in ~/gif for convenience. 

## Motivation and Context

The changes were based on feedback from rix users. 

## How Has This Been Tested?

This has been testing by executing the script and trying to add new cameras, change existing cameras, and seeing if gifs are created in the correct location and are able to be viewed using display. 

README.md has been updated for image_save and hdf5_to_gif.py

image

### DIFF
--- a/README.md
+++ b/README.md
@@ -258,20 +258,21 @@ usage: grep_ioc KEYWORD [hutch]<br/>
     <td>hdf5_to_gif.py</td>
     <td>
      Converts images in hdf5 files created with h5_img_collect to gifs.<br/>
-     Specify the path with -p and and how long each frame should last (ms) with -d.<br/>
-     Will save to cwd with as {original_filename}.gif<br/>
+     Specify the path with -p and and how long each frame should last (ms) with -t.<br/>
+     Will save to cwd or a specified directory with -d as {original_filename}.gif<br/>
     </td>
 </tr>
 
 <tr>
-    <td>image-saver.py</td>
+    <td>image_saver</td>
     <td>
      Uses h5_img_collect to save images from a camera in an hdf5 format.<br/>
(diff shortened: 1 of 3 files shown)


ZLLentz (approved):
This seems good to me
I don't have any pointed feedback, I think the feature extension and the minor cleanup (exceptions instead of exits) is good.

The one thing that gave me pause is using a generic directory name like `~/gifs` which conceivably could become a namespace collision with another application, even though it probably won't be.

KaushikMalapati:
How about ~/image-saver-gifs, or {--path}/image-saver-gifs?

KaushikMalapati:
> This seems good to me I don't have any pointed feedback, I think the feature extension and the minor cleanup (exceptions instead of exits) is good.
> 
> The one thing that gave me pause is using a generic directory name like `~/gifs` which conceivably could become a namespace collision with another application, even though it probably won't be.

How about ~/image-saver-gifs, or {--path}/image-saver-gifs?

ZLLentz:
I think all of those are better, but note that image-saver is also a very generic name!
It's ok to accept the gifs folder name and merge if we're unsure of what is best

KaushikMalapati:
Minor addition, you can now specify a path where gifs should be saved to. (Also allows you to take gui images without converting to gif.)

ZLLentz (approved):
I still think this looks good and have no specific feedback

PR #168, opened 2023-10-09
evrStatus pydm 

KaushikMalapati:
## Description

evrStatus now uses a pydm screen to display timing information. Additionally, the list of evr pvs has been updated to remove defunct pvs and add new ones.

## Motivation and Context

https://jira.slac.stanford.edu/browse/ECS-1023

## How Has This Been Tested?

shellcheck and launching the screen by specifying the hutch, pvs via command line arguments, and by entering manually using the -evr flag.

to_display comment has been updated 

image

### DIFF
--- a/scripts/evrStatus
+++ b/scripts/evrStatus
@@ -1,18 +1,26 @@
 #!/reg/g/pcds/pyps/conda/py39/envs/pcds-5.7.2/bin/python
+import subprocess
 from argparse import ArgumentParser as AP
 from sys import stdin
-import subprocess
-from qtpy import QtCore
-from qtpy.QtWidgets import (QLabel, QApplication, QTableWidget,
-                            QGridLayout, QHeaderView, QHBoxLayout)
-from pydm import Display
+
 from pydm.widgets import PyDMPushButton
 from pydm.widgets.label import PyDMLabel
+from qtpy import QtCore
+from qtpy.QtWidgets import (
+    QApplication,
+    QGridLayout,
+    QHBoxLayout,
+    QHeaderView,
+    QLabel,
+    QTableWidget,
+    QWidget,
+)
 
 
-class Expert(Display):
-    def __init__(self, pv):
+class Expert(QWidget):
+    def __init__(self, pv, parent=None):
         super().__init__()
+        self.parent = parent
         self.setWindowTitle(f"EVR Status Expert Mode ({pv})")
         self.app = QApplication.instance()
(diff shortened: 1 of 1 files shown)


scripts/evrStatus
```diff
@@ -132,47 +129,23 @@ req_evrs_func() {
 	    fi
         done
     fi
-    if [[ "$hutch" == "TMO" ]]; then
-        req_evrs+=("IM3K4:PPM:CAM:EVR,")       # This PV needs to be added manually b/c the name does not have "$hutch" identifier in it
```
ZLLentz, line 136: What's the story behind the various removals/additions to this section?

scripts/evrStatus
```diff
@@ -132,47 +129,23 @@ req_evrs_func() {
...
-    fi
-    local macros_exported=${macro_array[*]} # macros are exported to edm as a string
-    export macros_exported evr_num
-#   echo $macros_exported
+    source /cds/group/pcds/pyps/conda/pcds_conda
     echo "Displaying EVR status screen for " "$hutch" "..."    
     printf "\n"           
-    
-    /reg/g/pcds/epics-dev/screens/edm/common/current/evrStatus/edm-evrs.cmd
+    pydm /reg/g/pcds/epics-dev/screens/pydm/scripts/evr.py -p "${req_evrs[@]}" "${pv_args[@]}" &
```
ZLLentz, line 148: I see the main change here is shelling out to a dev evr.py screen- is there any way this can be brought in for review too?

It's really confusing having these be separate entities and version controlled separately. It was strange in the previous version as well, to be fair- so it doesn't have to be your scope to correct this.

Is evr.py used outside the scope of this script?

ZLLentz (commented):
I don't have any problems with this per se, but it seems kind of strange to me to have this shell script and to also have the python script and to keep them in separate locations.

For example, why is the cli handling largely done here? What is the function of this wrapper script, and why is it better to do these here instead of the python screen's script? Is there some function to this fractured setup or is largely historical?

scripts/evrStatus
KaushikMalapati: evr.py isn't used outside of this, I've integrated it

ZLLentz (approved):
I think this is pretty good, I like the high-level decisions you made here.
I included a few of my nitpicks and some general things to consider.
I can't review any further (e.g. running the script) until our servers come back online.

ZLLentz (approved):
I still think this looks good

Issue #147, opened 2023-03-01
Add a kinit helper to check when kinit needs to be run


ZLLentz:
## Expected Behavior

There should be a script you can run either standalone or inside other scripts to check whether your kerberos tickets are set up properly. This may also run kinit/aklog for you.

## Current Behavior

No such script exists

## Possible Solution

- Parse the output of klist
- Run klist -s and check for nonzero exit status

## Context

https://github.com/pcdshub/engineering_tools/pull/146#discussion_r1119272281

PR #169, opened 2023-10-11
Kinit helper

KaushikMalapati:
## Description

Adds two functions kauth and afsauth, which will create kerberos and afs tokens respectively. afsauth will call kauth. afsauth is called if the script is run, functions are accessible by sourcing. 

## Motivation and Context

#147 

## How Has This Been Tested?

shellcheck, running the script, and sourcing and trying functions individually and checking that tokens are created with klist and tokens or that correct error message is printed.

Code is commented

### DIFF
--- a/scripts/kinit_helper
+++ b/scripts/kinit_helper
@@ -9,6 +9,7 @@ function kauth {
             :
 	done
     fi
+    return 0
 }
 
 function afsauth {
@@ -18,14 +19,14 @@ function afsauth {
     if [[ $(hostname) != psbuild-rhel* ]]
     then
         echo "You must be on psbuild to create afs tokens"
-	exit
+	return 1
     fi
 
     # need to be in g-pcds afs grup
     if ! pts membership g-pcds 2>&1 | grep -q "$(whoami)"
     then
         echo "You do not have permission to use afs. See https://confluence.slac.stanford.edu/display/PCDS/Onboarding+Staff+Members"
-        exit
+        return 1
     fi
 
     # If token doesn't exist, create it
@@ -33,6 +34,7 @@ function afsauth {
     then	
         aklog
     fi
+    return 0
 }
 
 # if name == '__main__':
--- a/scripts/kinit_helper
+++ b/scripts/kinit_helper
@@ -4,10 +4,12 @@ function kauth {
     # If token doesn't exist, create it (will query for password)
     if ! klist -s
     then
-	while ! kinit
+	while ! kinit -l 365d -r 365d
 	do
             :
 	done
+    else
+	kinit -R &> /dev/null
     fi
     return 0
 }
@@ -15,7 +17,8 @@ function kauth {
 function afsauth {
     # afs needs kerberos token
     kauth
-    # only available on psbuild servers
+
+    # afs should be used from psbuild servers
     if [[ $(hostname) != psbuild-rhel* ]]
     then
         echo "You must be on psbuild to create afs tokens"
--- a/scripts/kinit_helper
+++ b/scripts/kinit_helper
@@ -1,0 +1,41 @@
+#!/bin/bash
+
+function kauth {
+    # If token doesn't exist, create it (will query for password)
+    if ! klist -s
+    then
+	while ! kinit
+	do
+            :
+	done
+    fi
+}
+
+function afsauth {
+    # afs needs kerberos token
+    kauth
+    # only available on psbuild servers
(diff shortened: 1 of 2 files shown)


scripts/kinit_helper
```diff
@@ -0,0 +1,43 @@
+#!/bin/bash
+
+function kauth {
+    # If token doesn't exist, create it (will query for password)
+    if ! klist -s
+    then
+	while ! kinit
+	do
```
ZLLentz, line 8: Does this loop just make it so you will be repeatedly queried until you type the password correctly? If so, this is pretty neat

scripts/kinit_helper
```diff
@@ -0,0 +1,43 @@
+#!/bin/bash
+
+function kauth {
```
ZLLentz, line 3: Consider the kinit cli args here.

Some useful options for this function:
`-l`: specify a lifetime (will be capped by the maximum allowed lifetime)
`-r`: specify a maximum renewal period (will be capped by the maximum allowed renewal period)
`-R`: lets you renew the existing ticket (if within the renewal period) instead of prompting for a password

I have a silly one-liner in my bashrc that may be of interest to explain why I like these:
```
klist -s >/dev/null 2>&1 && kinit -R || kinit -l 365d -r 365d
```
This is saying "check if we have valid tokens, if we do try to renew them, otherwise try to kinit with a lifetime of 1 year and a renewal period of 1 year" (comically large periods just to get the maximums)

So, when I already have the renewable token, this renews it for me without needing to type my password again.

I'm not 100% sure how this interacts with your workflows here, so I'm not making any specific recommendations, but here's some food for thought.

scripts/kinit_helper
```diff
@@ -0,0 +1,43 @@
...
+    return 0
+}
+
+function afsauth {
+    # afs needs kerberos token
+    kauth
+    # only available on psbuild servers
+    if [[ $(hostname) != psbuild-rhel* ]]
+    then
+        echo "You must be on psbuild to create afs tokens"
```
ZLLentz, line 21: This isn't quite true, but this is one of the few servers with the /afs mount point and this is where we should be doing the associated work, so the distinction is immaterial.

scripts/kinit_helper
```diff
@@ -0,0 +1,43 @@
...
+	return 1
+    fi
+
+    # need to be in g-pcds afs grup
+    if ! pts membership g-pcds 2>&1 | grep -q "$(whoami)"
```
ZLLentz, line 26: This is a neat trick, I didn't know about this utility

ZLLentz (dismissed):
Just some food for thought for potential improvements + random comments
I think the script is already useful as-is

scripts/kinit_helper
KaushikMalapati: Yes. shellcheck recommended using ! in a conditional instead of explicitly checking $?

KaushikMalapati:
Thanks for the feedback; I added renewing to kauth and changed the comment about psbuild

ZLLentz:
Reviewers can actually get a diff view of only the new changes since the last review, and they can also look at individual commits if they want to. Your changes look good to me, let me give this a try myself before putting out the approval again

scripts/kinit_helper
```diff
@@ -0,0 +1,46 @@
...
+	done
+    else
+	kinit -R &> /dev/null
+    fi
+    return 0
+}
+
+function afsauth {
+    # afs needs kerberos token
+    kauth
```
ZLLentz, line 19: I like this script, I'm probably going to be sourcing it and making use of both `kauth` and `afsauth` which are short, easy to remember function names.

Just one last tiny thing: the ordering of this step in `afsauth` and the ordering of the checks should probably be changed.

I think all the checks should happen prior to running `kauth` because `kauth` may prompt the user for their password. It's a bit annoying to type in your password, only afterwards to be told that you're on the wrong server.

I also think the group check should come before the hostname check, or both checks could be always be done. It'd just be nice for new team members to quickly be sent to the confluence page instead of sending them around to a different server first.

KaushikMalapati: Good idea. I switched the hostname and group checks and moved the kauth call so it would only be called if a token doesn't already exist.

ZLLentz:
One final thing: I completely forgot about the repo README (I should have brought that up earlier). If you add an entry to that for this script we're good to go.

ZLLentz (approved):
Ok great, thank you! Let's merge now

PR #173, opened 2024-02-14
ENH: set the default pydm protocol to ca, the most-used protocol

ZLLentz:
## Description

PyDM has an environment variable, `PYDM_DEFAULT_PROTOCOL`, that allows you to set a default protocol to use.
This PR sets the default protocol to `ca` (channel access (EPICS)), which is the most used protocol, allowing us to omit it when building out our screens. This also makes the conversion from edm to pydm smoother.

## Motivation and Context

When setting up a pydm channel, you're expected to provide a protocol in the form of a pseudo-url:
```
ca://MY:PVNAME
```

If, instead, you just use:
```
MY:PVNAME
```
You'll get the following error:
```
Channel MY:PVNAME did not specify a valid protocol and no default protocol is defined. This channel will receive no data. To specify a default protocol, set the PYDM_DEFAULT_PROTOCOL environment variable.
```

## How Has This Been Tested?

Setting this environment variable helped @patoppermann 's screen load properly (converted from edm)

n/a

### DIFF
--- a/scripts/dev_conda
+++ b/scripts/dev_conda
@@ -38,6 +38,7 @@ EPICS_ROOT="/cds/group/pcds/epics-dev"
 export PYTHONPATH="${APPS_ROOT}/dev/pythonpath"
 export PYQTDESIGNERPATH="${CONDA_ROOT}/dev_designer_plugins/:${PYQTDESIGNERPATH}"
 export PYDM_CONFIRM_QUIT=0
+export PYDM_DEFAULT_PROTOCOL=ca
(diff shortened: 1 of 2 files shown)


tangkong (approved):
This seems fine to me, I assumed and verified that this doesn't break other screens (in the off chance that this unintelligently appended `ca://` to the channel string)

Do we even use other channel types?...

ZLLentz:
> Do we even use other channel types?...

I have occasionally used `loc://` and have considered but decided against using `calc://`.
`pva://` exists and has been used as well, but not by me.

PR #174, opened 2024-02-23
motor-expert-screen checks for smaract-specific pvs so it works…

KaushikMalapati:
… with aliases without MCS2 in their prefix

## Description

Logic to select which edm screen to use now checks if LSCO_RBV pv is present, which is a smaract specific pv. 

## Motivation and Context

Some smaract axes have aliases without MCS2 in their prefix, so before they would use the default aerotech edm screen when launched from motor-expert-screen instead of a smaract screen.
https://jira.slac.stanford.edu/browse/ECS-4401

## How Has This Been Tested?

Launching motor-expert-screen with a smaract alias pv without mcs2 in its prefix and a smaract pv with mcs2 in its prefix and ensuring it uses a smaract screen instead of an aerotech screen.

A comment in the code has been updated

### DIFF
--- a/scripts/motor-expert-screen
+++ b/scripts/motor-expert-screen
@@ -96,17 +96,20 @@ elif [ "${rtyp}" == 'motor' ]; then
     caget "$PREFIX:PN" > /dev/null 2>&1
     if [ $? -eq 0 ]; then
         /reg/g/pcds/package/epics/3.14//modules/pcds_motion/${OLD_VERSION}/launch-motor.sh "$PREFIX" > /dev/null 2>&1 &
-    elif [[ $PREFIX == *"MCS2"* ]]; then # smaracts have "motor" rtyp, so we catch them with the PREFIX
-        enc=$(caget $PREFIX.UEIP)
-        if [[ $enc == *"Yes"* ]]; then
-            edm -x -eolc -m "MOTOR=$PREFIX" mcs2_main.edl > /dev/null 2>&1 &
-        else
-            edm -x -eolc -m "MOTOR=$PREFIX" mcs2_openloop.edl > /dev/null 2>&1 &
-        fi
     else
-        cd /reg/neh/home/klg/epics/ioc/common/aerotech/current/motorScreens
-        #cd /reg/neh/home4/mcbrowne/trunk2/ioc/common/aerotech/current/motorScreens
-        edm -x -eolc -m "MOTOR=${PREFIX}" ens_main.edl >& /dev/null &
+	caget "$PREFIX:LSCO_RBV" > /dev/null 2>&1
+	if [[ $? -eq 0 || $PREFIX == *"MCS2"* ]]; then # smaracts have "motor" rtyp, so we catch them with the PREFIX or with smaract specific pvs
+            enc=$(caget $PREFIX.UEIP)
+            if [[ $enc == *"Yes"* ]]; then
+		edm -x -eolc -m "MOTOR=$PREFIX" mcs2_main.edl > /dev/null 2>&1 &
+            else
+		edm -x -eolc -m "MOTOR=$PREFIX" mcs2_openloop.edl > /dev/null 2>&1 &
+            fi
+	else
+            cd /reg/neh/home/klg/epics/ioc/common/aerotech/current/motorScreens
+            #cd /reg/neh/home4/mcbrowne/trunk2/ioc/common/aerotech/current/motorScreens
+            edm -x -eolc -m "MOTOR=${PREFIX}" ens_main.edl >& /dev/null &
+	fi
     fi
 else
     edm -x -eolc -m "MOTOR=$PREFIX" ims_main.edl > /dev/null 2>&1 &
--- a/scripts/motor-expert-screen
+++ b/scripts/motor-expert-screen
@@ -97,8 +97,8 @@ elif [ "${rtyp}" == 'motor' ]; then
     if [ $? -eq 0 ]; then
         /reg/g/pcds/package/epics/3.14//modules/pcds_motion/${OLD_VERSION}/launch-motor.sh "$PREFIX" > /dev/null 2>&1 &
     else
-	caget "$PREFIX:LSCO_RBV" > /dev/null 2>&1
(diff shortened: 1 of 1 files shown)


scripts/motor-expert-screen
```diff
@@ -96,17 +96,20 @@ elif [ "${rtyp}" == 'motor' ]; then
...
-        if [[ $enc == *"Yes"* ]]; then
-            edm -x -eolc -m "MOTOR=$PREFIX" mcs2_main.edl > /dev/null 2>&1 &
-        else
-            edm -x -eolc -m "MOTOR=$PREFIX" mcs2_openloop.edl > /dev/null 2>&1 &
-        fi
     else
-        cd /reg/neh/home/klg/epics/ioc/common/aerotech/current/motorScreens
-        #cd /reg/neh/home4/mcbrowne/trunk2/ioc/common/aerotech/current/motorScreens
-        edm -x -eolc -m "MOTOR=${PREFIX}" ens_main.edl >& /dev/null &
+	caget "$PREFIX:LSCO_RBV" > /dev/null 2>&1
```
ZLLentz, line 100: Can you give an example of an IOC where there is a LSCO_RBV PV but not MCS2 in the PV name? Here's the result of my search:
```
zlentz@psbuild-rhel7-01:~$ grep_pv LSCO_RBV
/reg/d/iocData/ioc-lm1k2-mcs2-03/iocInfo
LM2K2:MCS2:03:m1:LSCO_RBV, "int64in"
LM2K2:MCS2:03:m2:LSCO_RBV, "int64in"

zlentz@psbuild-rhel7-01:~$
```
maybe grep_pv doesn't show aliases?

scripts/motor-expert-screen
ZLLentz, line 1: In general this approach is good if it is needed
This file is starting to get awkward though- we're getting upwards of 4 PVs in series, subjecting ourselves to potentially multiple CA timeouts. I have to wonder if a refactor is on the horizon.

scripts/motor-expert-screen
slactjohnson: @KaushikMalapati I think a better PV to use for this would be `PTYPE` or `PTYPE_RBV`. The `LSCO` PVs are recently added, and so IOCS based on older releases will not have this. The `PTYPE` PVs are specific to SmarAct, and have always been a part of the IOC.

scripts/motor-expert-screen
```diff
@@ -96,17 +96,20 @@ elif [ "${rtyp}" == 'motor' ]; then
...
+	# smaracts have "motor" rtyp, so we catch them with the PREFIX or with smaract specific pvs
+	if [[ $PREFIX == *"MCS2"* || $(caget "$PREFIX:PTYPE_RBV" > /dev/null 2>&1) -eq 0 ]]; then
+            enc=$(caget $PREFIX.UEIP)
+            if [[ $enc == *"Yes"* ]]; then
+		edm -x -eolc -m "MOTOR=$PREFIX" mcs2_main.edl > /dev/null 2>&1 &
+            else
+		edm -x -eolc -m "MOTOR=$PREFIX" mcs2_openloop.edl > /dev/null 2>&1 &
+            fi
+	else
+            cd /reg/neh/home/klg/epics/ioc/common/aerotech/current/motorScreens
```
slactjohnson, line 109: Woah, just noticed this. Why are we running screens out of Karl's (or Mike's) home directory?

scripts/motor-expert-screen
KaushikMalapati: I didn't know that PTYPE was specific to smaract, I've changed it to use PTYPE_RBV instead of LSCO_RBV

scripts/motor-expert-screen
KaushikMalapati: This showed up in the diff because I nested an existing conditional block. I could add /reg/g/pcds/epics/ioc/common/aerotech/R1.2.0/motorScreens to EDMDATAFILES and use ens_main.edl without changing directories. Is that reasonable?

slactjohnson: Moving to a released version is best, but that's a issue un-related to your PR. I'm still a little puzzled about the way this script decides to open the aerotech screen.  I'll think about this more, and perhaps open a separate issue on this repo.

ZLLentz: I agree, this is a problem but it's not a new problem

scripts/motor-expert-screen
```diff
@@ -96,17 +96,20 @@ elif [ "${rtyp}" == 'motor' ]; then
...
-            edm -x -eolc -m "MOTOR=$PREFIX" mcs2_main.edl > /dev/null 2>&1 &
-        else
-            edm -x -eolc -m "MOTOR=$PREFIX" mcs2_openloop.edl > /dev/null 2>&1 &
-        fi
     else
-        cd /reg/neh/home/klg/epics/ioc/common/aerotech/current/motorScreens
-        #cd /reg/neh/home4/mcbrowne/trunk2/ioc/common/aerotech/current/motorScreens
-        edm -x -eolc -m "MOTOR=${PREFIX}" ens_main.edl >& /dev/null &
+	# smaracts have "motor" rtyp, so we catch them with the PREFIX or with smaract specific pvs
+	if [[ $PREFIX == *"MCS2"* || $(caget "$PREFIX:PTYPE_RBV" > /dev/null 2>&1) -eq 0 ]]; then
```
ZLLentz, line 101: PTYPE_RBV has a ton of matching PVs that demonstrate the issue, so I think this is a useful pick

Somewhat unrelated: does bash skip evaluation of the second part of the or statement if it can?

ZLLentz:
I'm going to run this myself a few times and then approve

ZLLentz (approved):
Code looks good and it works for the random sample of motors I've checked :+1:

scripts/motor-expert-screen
slactjohnson: I made a new issue for this here: https://github.com/pcdshub/engineering_tools/issues/175

scripts/motor-expert-screen
KaushikMalapati: Bash short-circuits && and || or statements - https://www.gnu.org/software/bash/manual/bash.html#Lists

PR #178, opened 2024-04-04
Add in new script for resuming PVs in the archiver.

slactjohnson:
## Description
Adds in a script for resuming PVs in the archiver from the command line. 

## Motivation and Context
TMO found a PV that was paused in the archiver and the people online at the time did not know how to fix it. The _did_ know how to check the status using `archive-status`, so I thought perhaps a similar bash tool would be helpful in the future. 

https://slac.slack.com/archives/CUC9NLU8H/p1712197816806739

## How Has This Been Tested?
Tested live using a PV in the FTL. Paused the pv manually from the archiver web interface. 

image

### DIFF
--- a/scripts/archive-resume
+++ b/scripts/archive-resume
@@ -1,0 +1,24 @@
+#!/bin/bash
+usage()
+{
+cat << EOF
+usage: $0 <PV>
+
+Resume archiving of a PV in the archiver, e.g. $0 GDET:FEE1:241:ENRC
+EOF
+}
+
+if [[ -z $1 ]]; then
+    usage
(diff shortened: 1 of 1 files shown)


slactjohnson:
This script is almost identical to `archive-status`, it just uses a different BPL. In principle we could make this a more general bash script that supports more bpls, but.... I think that's a problem for another day.

slactjohnson:
Ping @silkenelson @tongju12

tongju12 (approved):
Sorry, Taylor, I totally forgot this.
It looks good to me. Thanks!

PR #183, opened 2024-04-24
MNT/STY/TST: Add pre-commit hooks, CI job, standardize style

tangkong:
## Description
- adds pre-commit hooks and relevant .rc files to the repository
- adds github actions pre-commit job to run on pull requests
- (WIP) apply changes suggested by pre-commit (flake8, shellcheck, etc)
  - [x] python files 
~- [ ] bash scripts (hooooo boy)~
- Converts some old py2 scripts to py3

## Motivation and Context
The style is whack.

Shellcheck is a much larger effort, and one that we should be careful with.  We'll have to go through those more carefully, but for now the CI will only check diff'd files

I spent way too much time trying to grab specific sha's for pre-commit's `--from-ref` and `--to-ref` arguments, but the location of that information in the `github` payload was inconsistent between push and pull-request.  At that point I figured I should just get it working.  

## How Has This Been Tested?
Pre-commit vomits all over my terminal repeatedly

This PR

### DIFF
--- a/scripts/get_info.py
+++ b/scripts/get_info.py
@@ -1,7 +1,10 @@
 import argparse
-import socket
+import logging
 import os
+import socket
 import sys
+
+import requests
 
 parser = argparse.ArgumentParser()
 parser.add_argument("--run", help="get last run", action='store_true')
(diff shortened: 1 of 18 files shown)


scripts/get_info.py
```diff
@@ -12,91 +15,98 @@
...
-                    hutch=ihutch.upper()
-                    foundHutch=True
+                if path.find(ihutch) >= 0:
+                    hutch = ihutch.upper()
+                    foundHutch = True
                     break
-            if not foundHutch and path.find('xrt')+hostname.find('xrt')>=-1 or path.find('xtod')+hostname.find('xtod')>=-1:
-                hutch='LFE' #because we have so many names for the same subnet.
-                foundHutch=True
+            if (not foundHutch and path.find('xrt')+hostname.find('xrt') >= -1
```
ZLLentz, line 104: I'm slightly surprised this line passes flake8 with the lack of + spacing

ZLLentz (approved):
I liked the specific implementation of the pre-commit a lot
The specific changes all look correct/nondestructive to me

scripts/get_info.py
tangkong: I'm also confused about this...

Issue #182, opened 2024-04-23
ENH: Add pre-commit checks, basic CI


tangkong:
## Expected Behavior
We should add pre-commit checks (ruff, flake8, shellcheck) to this repo
We should make at least those checks run as CI

## Current Behavior
No checks

## Context
Brought up in #180

## Your Environment
pcds-5.8.2

PR #184, opened 2024-05-06
MNT: Shellchecking bash scripts

KaushikMalapati:
## Description

Very minor edits that don't change anything important.

I shellchecked the bash scripts in /scripts (making edits based on warnings, errors and info messages that shellcheck gave). Some notes and questions I have are:

- Why is wc needed on line 319 of camviewer and what is this line supposed to do?
- Line 502 of makepeds_psana should be an elif unless xtcav_dark can change XTCAV?
- In line 196 of ioctool, the third if condition seemed incomplete and I tried to fix it by matching the regex to the string in question
- In these three places I arbitrarily chose to follow the rix soure, but I could choose different ones or just ignore the warnings
          stopdaq:    # shellcheck source=/reg/g/pcds/dist/pds/rix/scripts/setup_env.sh
          restartdaq:    # shellcheck source=/reg/g/pcds/dist/pds/rix/scripts/setup_env.sh
          checkcnf:    # shellcheck source=/reg/g/pcds/dist/pds/rix/scripts/setup_env.sh
- I generally opted to use double brackets in if conditions to avoid word splitting warnings, but I could have quoted instead. I don't think we would have any portability concerns about this but I wanted to at least mention it
- I added the r switch to all instances of read to prevent backslash mangling in user input - I did not see a case in any of these scripts where someone would want to do this intentionally (SC2162  for context)
- Some of the most common edits - Double quoting variables, removing extraneous dollar signs in arithmetic expressions, fixing the order of backticks and braces for awk, replacing legacy backticks '' with `$()` for commands, using -gt instead of >, using grep -c or pgrep instead of grep | wc -l and ps | grep, and checking exit codes directly with if cmd; or if !cmd instead of cmd; if [$?]
- xpp_update_happi_line, grep_pv, ioctool, makepeds_psana, and motor-expert_screen all have warning SC2164 about cd failing. Should this be added to .shellcheckrc to always ignore it? Usually, the directories are sourced, or are constant paths that exist, or are checked explicitly beforehand
- configdb_readxtc, wherepsana, startami, motorInfo, restartdaq, makepeds, and motor-expert-screen (double-dipper) all have warning SC2034 about unused variables. I don't think this warning should be globally suppressed, but I'm also not sure when if the variables are used elsewhere and should exported or if they are no longer being used and can be removed.

I also found a tool called shfmt which we could add a precommit job for. Here is an example of the diff it would print for kinit_helper

```
-function kauth {
+function kauth
+{
     # If token doesn't exist, create it (will query for password)
-    if ! klist -s
-    then
-       while ! kinit -l 365d -r 365d
-       do
+    if ! klist -s; then
+        while ! kinit -l 365d -r 365d; do
             :
-       done
+        done
     else
-       kinit -R &> /dev/null
+        kinit -R &> /dev/null
     fi
     return 0
 }

-function afsauth {
+function afsauth
+{
     # need to be in g-pcds afs grup
-    if ! pts membership g-pcds 2>&1 | grep -q "$(whoami)"
-    then
+    if ! pts membership g-pcds 2>&1 | grep -q "$(whoami)"; then
         echo "You do not have permission to use afs. See https://confluence.slac.stanford.edu/display/PCDS/Onboarding+Staff+Members"
         return 1
     fi

     # afs should be used from psbuild servers
-    if [[ $(hostname) != psbuild-rhel* ]]
-    then
+    if [[ $(hostname) != psbuild-rhel* ]]; then
         echo "You must be on psbuild to create afs tokens"
-       return 1
+        return 1
     fi

     # If token doesn't exist, create it
-    if ! tokens | grep -q $UID
-    then
+    if ! tokens | grep -q $UID; then
         # afs needs kerberos token
-       kauth
-       aklog
+        kauth
+        aklog
     fi
     return 0
 }

 # if name == '__main__':
-if [[ "${BASH_SOURCE[0]}" == "${0}" ]]; then
+if [[ ${BASH_SOURCE[0]} == "${0}" ]]; then
     afsauth
 fi
```

It basically formats things nicely. I have the indenting set to four spaces right now, but it could be a different number or tabs instead. I haven't formatted anything yet because I don't want to add this unilaterally,

## Motivation and Context

#182 
To standardize style and fix existing shellcheck issues so that future contributors only see pre-commit/workflow errors related to their changes instead of what was already there.

## How Has This Been Tested?

Interactively 

N/A


KaushikMalapati:
I will turn this into a real pr once I test it a bit more.

tangkong:
Is the way to go here adding shellcheck AND shellfmt to precommit?

KaushikMalapati:
> Is the way to go here adding shellcheck AND shellfmt to precommit?

Yes. There's some overlap but shellcheck would be for catching errors and shfmt would be for consistent styling.

ZLLentz:
> Very minor edits

> 46 files

First glance looks great but we're going to need to review and test carefully

ZLLentz:
I think `shellfmt` and similar sorts of auto-formatting tools are great
As long as it doesn't disagree with shellcheck on anything we can include it

scripts/camViewer
ZLLentz, line 1: I'd flag this one as important to test thoroughly based on the number of changes

scripts/checkcnf
```diff
@@ -18,6 +18,7 @@ HUTCH=$(get_info --gethutch)
 
 LCLS2_HUTCHES="rix, tmo, ued"
 if echo "$LCLS2_HUTCHES" | grep -iw "$HUTCH" > /dev/null; then
+    # shellcheck source=/reg/g/pcds/dist/pds/rix/scripts/setup_env.sh
```
ZLLentz, line 21: I think it's totally reasonable to "just pick one" here and elsewhere, and rix is as good as any other.

scripts/get_curr_exp
ZLLentz, line 1: This one should also be tested thoroughly because it gets used a lot, even though the changes are minor

scripts/camViewer
```diff
@@ -320,17 +315,18 @@ name_to_pv "$CAMNAME" "$hutch"
 if [ $MAINSCREEN -gt 0 ]; then
     tmp=/tmp/cv.$$
     list_cams "$hutch" >$tmp
-    c=$(wc <$tmp | grep "$CAMNAME" $tmp | awk '{print $1}')
+
+    c=$(wc $tmp | grep "$CAMNAME" $tmp | awk '{print $1}') # not sure what is happening here
```
ZLLentz, line 319: This wc is also a mystery to me, I think it must be an error. This will always land in the "camera name not found in hutch" branch, which I've noticed when using the script!

scripts/ioctool
```diff
@@ -194,22 +193,22 @@ if [[ $NAME == *':'* ]]; then #we are assuming that PVs have a colon in them
 #2) If so, then search for ioc using grep_ioc
 #3) If that returns nothing, try converting the camera name to a PV and search for that
 #4) If neither of those produce results return "Did not find name anywhere"
-elif [[ $NAME == *'-'* || $NAME == *'_'* || ^[[:alnum:]]+$ ]]; then 
+elif [[ $NAME == *'-'* || $NAME == *'_'* || $NAME =~ ^[[:alnum:]]+$ ]]; then
```
ZLLentz, line 196: I think your update here makes sense

scripts/makepeds
ZLLentz, line 1: RE: unused variable warnings: does it still give the warning on variables that are exported but not used? If it doesn't warn for exported variables, then I think this warning is always useful, since in the shell we need to be explicit about our exports for the variables to be used elsewhere.

scripts/makepeds_psana
ZLLentz, line 1: Flagging as another important script to test

scripts/takepeds
```diff
@@ -10,7 +10,7 @@ Takes a run with dark images for use in pedestals, and posts to the elog.
 EOF
 }
 
-if [[($1 == "--help") || ($1 == "-h")]]; then
+if [[ ($1 == "--help") || ($1 == "-h")]]; then
```
ZLLentz, line 13: Maybe missing space here
```suggestion
if [[ ($1 == "--help") || ($1 == "-h") ]]; then
```

ZLLentz (commented):
I did my best to review each file line-by-line and make sure I agree with every individual change.
I also did my best to answer questions from your description when I was able.
I think it's appropriate for at least one more person to review and to get some help testing some of these.

tangkong (commented):
I'll admit my eyes started to glaze over a bit but I did read every file.  (I think)

I'll spend time trying these out where I can.

scripts/wherepsana
```diff
@@ -63,40 +63,39 @@ if [[ -z $CONFIG ]]; then
...
+    for ((idx=0; idx<${#PSANA_NODES[@]}; ++idx)); do
+        NODESTR=$NODESTR${PSANA_NODES[idx]}',';
     done
-    echo ${NODESTR::-1}
+    echo "${NODESTR::-1}"
 else
-    EVENT=(`grep event /tmp/procServStatus.$$ | awk {'print $1'}`)
-    PSANA_MASKS=(`grep monreqsrvpsana /tmp/procServStatus.$$ | awk -F "-i " {'print $2'} | awk {'print $1'}`)
-    EVENT_MASKS=(`grep event /tmp/procServStatus.$$ | awk -F "-s " '{print $2;}' | awk '{print lshift(1,$1);}'`)
+    mapfile -t EVENT < <(grep event /tmp/procServStatus.$$ | awk '{print $1}')
```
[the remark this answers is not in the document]
KaushikMalapati: SC2207,  to avoid globbing/splitting.

scripts/new_bare_repo
```diff
@@ -43,7 +43,7 @@ elif [[ ${CWD} != *${REPO}* ]]; then
 else
     git init
     cp -u /cds/sw/tools/bin/eco_tools/gitignore.template .gitignore
-    git add *.cfg Makefile .gitignore
+    git add -- *.cfg Makefile .gitignore
```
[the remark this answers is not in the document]
KaushikMalapati: SC2035 (info): Use ./*glob* or -- *glob* so names with dashes won't become options.
Realistically, I don't think this would ever cause an issue here so we could add it to .shellcheckrc

scripts/makepeds
KaushikMalapati: There won't be a warning if you export the variable (SC2034)

scripts/daq_control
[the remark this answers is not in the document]
KaushikMalapati: Three years, probably can delete this.

scripts/camViewer
ZLLentz: So, revisiting here, probably we should be grepping through the config file instead of word-counting the temp file? This is definitely a pre-existing bug though and you're not obligated to mix this sort of thing with a style PR

silkenelson: There is no 'the' config file - the config file can have includes and it's not clear to me that we want to write parsing for this.

ZLLentz: Are you recommending removing this check altogether?

ZLLentz: Looking more closely: the `list_cams` function here already does a half-trying version of working through the includes (hard-coded).
So, the quick fix is to the simply remove the wc clause and start from the grep, but whether or not this is a thing we should do at all is another question.

PR #187, opened 2024-07-02
MAINT: remove glx env var, no longer needed and causing issues

ZLLentz:
## Description

Remove opengl-related environment variable from the conda startup scripts

## Motivation and Context

At LCLS, this was a temporary workaround that is causing a new issue.
- This variable is causing https://jira.slac.stanford.edu/browse/ECS-5700
- This variable was originally a workaround for https://jira.slac.stanford.edu/browse/ECS-1867, see https://jira.slac.stanford.edu/browse/ECS-1878, https://github.com/pcdshub/engineering_tools/pull/132, https://github.com/pcdshub/engineering_tools/pull/133
- Interactively, my applications no longer crash without this variable set: previously, even simply starting `pydm` would break everything.
- Now, sometimes I get a new OpenGL warning but no crashes.

In general, forcing the renderer settings in the conda environment sourceable led to a very confusing bug. This should have been handled (and seemingly has been handled) by adjusting the server settings.

I'm not sure why this variable is causing crashes now, but I am sure that I don't want to be in charge of messing with these sorts of settings in the future.

## How Has This Been Tested?

Interactively only, as me

Only here and on Jira in https://jira.slac.stanford.edu/browse/ECS-5700

### DIFF
--- a/scripts/dev_conda
+++ b/scripts/dev_conda
@@ -55,13 +55,5 @@ source "${TOKEN_DIR}/typhos.sh"
 # The same was done with PyDM and this fixes Designer and friends.
 if [ -n "${SSH_CONNECTION}" ]; then
   export QT_QUICK_BACKEND="software"
-  # Check for errors in the opengl rendering
-  # Manually fall back to mesa library if we know there is a problem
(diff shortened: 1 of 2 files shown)


tangkong (approved):
🤷  This should really be outside of our wheelhouse, and this gets us closer to that

PR #189, opened 2024-07-17
ENH: add quick script for repointing afs remotes to github

ZLLentz:
## Description

Add `afs-remote-fix` (`afs_remote_fix.py`), a tiny script for changing a user's afs remote to point toward their fork and to pcdshub.

## Motivation and Context

We're migrating the afs repos so this will be useful.

## How Has This Been Tested?

Interactively only

Here only in the readme and in the file

### DIFF
--- a/scripts/afs_remote_fix.py
+++ b/scripts/afs_remote_fix.py
@@ -1,0 +1,50 @@
+#!/usr/bin/python3
+"""
+afs-remote-fix is a script for repointing your afs git remotes towards github.
+
+It will check if you're currently in a git ioc repo with an afs remote as your origin,
+switching it to your fork and adding an upstream remote to pcdshub.
+
+An afs remote is a git remote with a path that looks something like this:
+/afs/slac.stanford.edu/g/cd/swe/git/repos/package/epics/ioc/common/ims.git
+
+If this path is your origin, after running the script you will instead have:
+
+origin          git@github.com:your-username/ioc-common-ims.git
+upstream        git@github.com:pcdshub/ioc-common-ims.git
+
+If the origin is not an afs path, this script will exit without taking any action.
+
+The script will prompt you for your github username, which is not necessarily
+the same as your slac username.
+"""
+import subprocess
+
+
+def main() -> int:
+    origin = subprocess.check_output(["git", "remote", "get-url", "origin"], universal_newlines=True).strip()
+    if not origin.startswith("/afs/"):
+        print(f"Your origin {origin} is not an afs repo.")
+        return 1
+    elif "/ioc/" not in origin:
+        print(f"Your origin {origin} is not an ioc repo.")
+        return 1
+    _, ioc_path = origin.split("/ioc/")
+    repo_name = f"ioc-{ioc_path.replace('/', '-', 1)}"
+    username = input("Please input your github username:\n")
+    new_origin = f"git@github.com:{username}/{repo_name}"
+    new_upstream = f"git@github.com:pcdshub/{repo_name}"
+    print("\nPlanning to set:")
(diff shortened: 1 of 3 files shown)


scripts/afs_remote_fix.py
```diff
@@ -0,0 +1,53 @@
...
+    _, ioc_path = origin.split("/ioc/")
+    repo_name = f"ioc-{ioc_path.replace('/', '-', 1)}"
+    username = input("Please input your github username:\n")
+    new_origin = f"git@github.com:{username}/{repo_name}"
+    new_upstream = f"git@github.com:pcdshub/{repo_name}"
+    print("\nPlanning to set:")
+    print(f"origin to {new_origin}")
+    print(f"upstream to {new_upstream}")
+    confirm = input("Confirm? y/n\n")
+    if not confirm.lower().startswith("y"):
```
tangkong, line 44: Now that `click` is in pcds-conda, `click.confirm` is a thing https://click.palletsprojects.com/en/8.1.x/prompts/#confirmation-prompts

scripts/afs_remote_fix.py
```diff
@@ -0,0 +1,53 @@
...
+        return 1
+    subprocess.run(["git", "remote", "set-url", "origin", new_origin])
```
tangkong, line 46: Are we certain everyone has a remote called origin?  Maybe the more consistent thing to do is remove all the remotes then create `origin` and `upstream`

tangkong (commented):
This looks largely good to me

scripts/afs_remote_fix.py
ZLLentz: You're right, if I switch this to use pcds_conda click is better.

scripts/afs_remote_fix.py
ZLLentz: You're right, this is a big assumption. This is definitely what you get if you just `git clone` from afs. Let me check what happens when you use `epics-checkout`.

ZLLentz: Here's the wild structure you get from `epics-checkout`:
```
zlentz@psbuild-rhel7-03:~/.../ims/ims-git(master +)$ pwd
/cds/home/z/zlentz/temp/afs_ioc_checks/ims/ims-git
zlentz@psbuild-rhel7-03:~/.../ims/ims-git(master +)$ git remote -v
origin  /afs/slac/g/cd/swe/git/repos/package/epics/ioc/common/ims.git (fetch)
origin  /afs/slac/g/cd/swe/git/repos/package/epics/ioc/common/ims.git (push)
```

scripts/afs_remote_fix.py
tangkong: I realize I'm just always using pcds-conda so it's not really a time loss for me.

scripts/afs_remote_fix.py
tangkong: I guess if someone was savvy enough to change their remote names from the default, they're probably savvy enough to fix this without the help of the script?

ZLLentz: That's my current thought too, but I'm biased by not wanting to do more work on this. I'm happy to change the behavior if it would be helpful.

tangkong: I think this is good enough for the purpose.  probably more important to get this out there for the migration

tangkong (approved):

ZLLentz:
I'll leave these up for a bit longer to allow for other reviews

ZLLentz:
If someone comes here later from slack: please still do a post-merge review if you'd like, we can change these scripts as needed.

PR #190, opened 2024-07-23
FIX/ENH: fix ioc-deploy auth, provide ssh-agent-helper

ZLLentz:
## Description

- [x] Switch the http checks in `ioc-deploy` to git clone checks using ssh auth.
- [x] Provide `ssh-agent-helper` for using the ssh-agent based on https://github.com/pcdshub/twincat-bsd-ansible/blob/master/scripts/ssh_agent_helper.sh
- [x] Warn users that they might want to try the `ssh-agent-helper`

## Motivation and Context

- `ioc-deploy` currently only works on public repos because it receives 404s from http checks of internal and private repos for logged out users. If we instead use the user's shell's existing ssh authentication, we don't need to pass passwords around, distribute PATs, etc.
- I'd like to provide some ssh-agent helpers so that people only have to type in their ssh key one time. It's surprisingly nuanced to set these up, it's best practice to put time limits on your ssh-adds and make sure to clean up the agent on shell exit. The result is pretty concise but those are hard-fought lines of bash script...

## How Has This Been Tested?

Interactively only

Only here, so far


ZLLentz:
I chatted with Robert and our consensus is that this probably shouldn't set up SSH keys, but it should let you know that your agent isn't set up and suggest actions to fix this.

slactjohnson (dismissed):
No comments, LGTM

scripts/ioc_deploy.py
```diff
@@ -313,6 +322,32 @@ def get_version() -> str:
...
+    cmd = ["git", "clone", f"git@github.com:{github_org}/{name}", "--depth", "1"]
+    if release:
+        cmd.extend(["-b", release])
+    if target_dir:
+        cmd.append(target_dir)
+    kwds = {"check": True}
+    if working_dir:
+        kwds["cwd"] = working_dir
+    if not verbose:
+        kwds["stdin"] = subprocess.PIPE
```
tangkong, line 345: Why are we redirecting stdin here?  Do we mean to redirect stdout?

scripts/ssh-agent-helper
```diff
@@ -0,0 +1,31 @@
...
+if [ "$rval" -eq 2 ]; then
+    echo "Starting ssh agent"
+    eval "$(ssh-agent -s)" &> /dev/null
+    ssh_agent_helper_cleanup() {
+        if [ -n "${SSH_AGENT_PID}" ]; then
+            echo "Cleaning up SSH agent"
+            eval "$(ssh-agent -k)" &> /dev/null
+        fi
+    }
+    trap ssh_agent_helper_cleanup EXIT
```
tangkong, line 25: This is some new bash-fu to me.

The idea here is that we start the ssh-agent, then kill it if `SSH_AGENT_PID` doesn't exist at the end of this script?  Doesn't the agent need to be running for keys to be applied?

What sets the `SSH_AGENT_PID` env var?  I don't have it in any of my sessions, with a forwarded agent at least.

scripts/ioc_deploy.py
ZLLentz: whoops, yes
I guess all the important stuff to redirect was in stderr because I didn't notice anything wrong

scripts/ssh-agent-helper
ZLLentz: This `trap` line will call `ssh_agent_helper_cleanup` when the user's shell exits (the shell that sources this script). The intent of that function is to kill the ssh agent when the user logs out, because otherwise the agent process persists on the server indefinitely.

`SSH_AGENT_PID` is set by `eval "$(ssh-agent -s)` when you run it on the local host, it doesn't exist if you forward an agent. When you call `ssh-agent -s` it dumps a bunch of environment variable set commands to stdout. It's really weird but that's how it's coded. Then, you `eval` this output to get the environment variables that `ssh` needs to properly use the agent for your key.

If the `SSH_AGENT_PID` is no longer set, it's because someone already cleaned up the agent. In fact, without this environment variable set, the agent kill command doesn't know which agent to kill.

Some of this comment should be added to the script for clarity/documentation purposes.

tangkong: oh weird.  the man page for `ssh-agent` doesn't mention any env vars in the `-s` option documentation, and I'd never expect a command to just dump other commands to stdout 😕

ZLLentz: It's sort of mentioned in the man page but not in a very clear way. It doesn't expand much on what it's really doing.
```
There are two main ways to get an agent set up: The first is that the agent starts a new subcommand into which some environment variables are exported, eg ssh-agent xterm &.  The second is that the agent prints the needed shell commands (either sh(1) or csh(1) syntax can
be generated) which can be evaluated in the calling shell, eg eval `ssh-agent -s` for Bourne-type shells such as sh(1) or ksh(1) and eval `ssh-agent -c` for csh(1) and derivatives.
```

ZLLentz:
I fixed the stdout/stdin typo and greatly expanded the comments in ssh-agent-helper so it can be useful as a primer on how the agent setup works.

ZLLentz:
One more to-do: I need to add `ssh-agent-helper` to the README

tangkong (approved):
LGTM, thanks for explaining the bash shenanigans to me

janeliu-slac:
Sorry, I didn't have much to contribute. Read through the file changes and brushing up on bash scripting.

PR #195, opened 2024-08-28
WIP: add permission handling to ioc-deploy

ZLLentz:
## Description

WIP
https://jira.slac.stanford.edu/browse/ECS-6004

### DIFF
--- a/scripts/ioc_deploy.py
+++ b/scripts/ioc_deploy.py
@@ -3,25 +3,30 @@
 ioc-deploy is a script for building and deploying ioc tags from github.
 It will create a shallow clone of your IOC in the standard release area at the correct path and "make" it.
 If the tag directory already exists, the script will exit.
+After making the IOC, we'll write-protect all files and all directories that contain built files
+(e.g. those that contain files that are not tracked in git).
+We'll also write-protect the top-level directory to help indicate completion.
 
 Example command:
 "ioc-deploy -n ioc-foo-bar -r R1.0.0"
 
 This will clone the repository to the default ioc directory and run make
-using the currently set EPICS environment variables.
+using the currently set EPICS environment variables, then apply write protection.
 
 With default settings this will clone
 
 from https://github.com/pcdshub/ioc-foo-bar
 to /cds/group/pcds/epics/ioc/foo/bar/R1.0.0
 
-then cd and make.
+then cd and make and chmod as appropriate.
 """
 
 import argparse
 import enum
 import logging
 import os
+import os.path
+import stat
 import subprocess
 import sys
 from pathlib import Path
@@ -46,6 +51,8 @@ if sys.version_info >= (3, 7, 0):
         release: str = ""
         ioc_dir: str = ""
         github_org: str = ""
+        remove_write_protection: bool = False,
+        apply_write_protection: bool = False,
         auto_confirm: bool = False
         dry_run: bool = False
         verbose: bool = False
@@ -92,6 +99,16 @@ def get_parser() -> argparse.ArgumentParser:
         help=f"The github org to deploy IOCs from. This defaults to $GITHUB_ORG, or {GITHUB_ORG_DEFAULT} if the environment variable is not set. With your current environment variables, this defaults to {current_default_org}.",
     )
     parser.add_argument(
+        "--remove-write-protection",
+        action="store_true",
+        help="If provided, instead of doing a release, we will remove write protection from an existing release. Incompatible with --apply-write-protection."
+    )
+    parser.add_argument(
+        "--apply-write-protection",
+        action="store_true",
+        help="If provided, instead of doing a release, we will add write protection to an existing release. Incompatible with --remove-write-protection."
+    )
+    parser.add_argument(
         "--auto-confirm",
         "--confirm",
         "--yes",
@@ -126,12 +143,17 @@ def main(args: CliArgs) -> int:
 
     Will either return an int return code or raise.
     """
-    if not (args.name and args.release):
-        raise ValueError(
-            "Must provide both --name and --release. Check ioc-deploy --help for usage."
-        )
-
     logger.info("Running ioc-deploy: checking inputs")
+    has_input_error = False
+    if not (args.name and args.release):
+        logger.error("Must provide both --name and --release. Check ioc-deploy --help for usage.")
+        has_input_error = True
+    if args.apply_write_protection and args.remove_write_protection:
+        logger.error("Must provide at most one of --apply-write-protection and --remove-write-protection")
+        has_input_error = True
+    if has_input_error:
+        return ReturnCode.EXCEPTION
+
     upd_name = finalize_name(
         name=args.name, github_org=args.github_org, verbose=args.verbose
     )
@@ -143,6 +165,23 @@ def main(args: CliArgs) -> int:
     )
     deploy_dir = get_target_dir(name=upd_name, ioc_dir=args.ioc_dir, release=upd_rel)
 
+    # Branch off for permission modification options
+    if args.apply_write_protection:
+        logger.info(f"Applying write permissions to {deploy_dir}")
+        if not args.auto_confirm:
+            user_text = input("Confirm target? y/n\n")
+            if not user_text.strip().lower().startswith("y"):
+                return ReturnCode.NO_CONFIRM
+        return set_permissions(deploy_dir=deploy_dir, protect=True, dry_run=args.dry_run)
+    elif args.remove_write_protection:
+        logger.info(f"Removing write permissions from {deploy_dir}")
+        if not args.auto_confirm:
+            user_text = input("Confirm target? y/n\n")
+            if not user_text.strip().lower().startswith("y"):
+                return ReturnCode.NO_CONFIRM
+        return set_permissions(deploy_dir=deploy_dir, protect=False, dry_run=args.dry_run)
+
+    # Main deploy routines
     logger.info(f"Deploying {args.github_org}/{upd_name} at {upd_rel} to {deploy_dir}")
     if Path(deploy_dir).exists():
         raise RuntimeError(f"Deploy directory {deploy_dir} already exists! Aborting.")
@@ -167,6 +206,10 @@ def main(args: CliArgs) -> int:
     rval = make_in(deploy_dir=deploy_dir, dry_run=args.dry_run)
     if rval != ReturnCode.SUCCESS:
         logger.error(f"Nonzero return value {rval} from make")
+        return rval
+    rval = set_permissions(deploy_dir=deploy_dir, protect=True, dry_run=args.dry_run)
+    if rval != ReturnCode.SUCCESS:
+        logger.error(f"Nonzero return value {rval} from set_permissions")
         return rval
     return ReturnCode.SUCCESS
 
@@ -299,6 +342,86 @@ def make_in(deploy_dir: str, dry_run: bool) -> int:
         return ReturnCode.SUCCESS
     else:
         return subprocess.run(["make"], cwd=deploy_dir).returncode
+
+
+def set_permissions(deploy_dir: str, protect: bool, dry_run: bool) -> int:
+    """
+    Apply or remove write protection from a deploy repo.
+
+    You may or may not have permissions to do this to releases created by other users.
+
+    Applying write permissions (protect=True) involves removing the "w" permissions
+    from all files, and from directories that contain untracked files.
+    We will also remove permissions from the top-level direcotry.
+
+    Removing write permissions (protect=False) involves adding "w" permissions
+    for the owner and for the group. "w" permissions will never be added for other users.
+    We will also add write permissions to the top-level directory.
+    """
+    if not protect:
+        # Lazy and simple: chmod everything
+        perms = get_add_write_rule(os.stat(deploy_dir, follow_symlinks=False).st_mode)
+        os.chmod(deploy_dir, perms)
+        for dirpath, dirnames, filenames in os.walk(deploy_dir):
+            for name in dirnames + filenames:
+                full_path = os.path.join(dirpath, name)
+                perms = get_add_write_rule(os.stat(full_path, follow_symlinks=False).st_mode)
+                os.chmod(full_path, perms)
+        return ReturnCode.SUCCESS
+
+    # Compare the files that exist to the files that are tracked by git
+    try:
+        ls_files_output = subprocess.check_output(
+            ["git", "-C", deploy_dir, "ls-files"],
+            universal_newlines=True,
+        )
+    except subprocess.CalledProcessError as exc:
+        return exc.returncode
+    deploy_path = Path(deploy_dir)
+    tracked_paths = [deploy_path / subpath.strip() for subpath in ls_files_output.splitlines()]
+    build_dir_paths = set()
+
+    def accumulate_subpaths(subpath: Path):
+        for path in subpath.iterdir():
+            if path.is_symlink():
+                continue
+            elif path.is_dir():
+                accumulate_subpaths(path)
+            elif path.is_file():
+                if path not in tracked_paths:
+                    build_dir_paths.add(str(path.parent))
+
+    accumulate_subpaths(deploy_path)
+
+    # Follow the write protection rules from the docstring
+    perms = get_remove_write_rule(os.stat(deploy_dir, follow_symlinks=False).st_mode)
+    os.chmod(deploy_dir, perms)
+    for dirpath, dirnames, filenames in os.walk(deploy_dir):
+        for dirn in dirnames:
+            if dirn in build_dir_paths:
+                full_path = os.path.join(dirpath, dirn)
+                perms = get_remove_write_rule(os.stat(full_path, follow_symlinks=False).st_mode)
+                os.chmod(full_path, perms)
+        for filn in filenames:
+            full_path = os.path.join(dirpath, filn)
+            perms = get_remove_write_rule(os.stat(full_path, follow_symlinks=False).st_mode)
+            os.chmod(full_path, perms)
+
+    return ReturnCode.SUCCESS
+
+
+def get_remove_write_rule(perms: int) -> int:
+    """
+    Given some existing file permissions, return the same permissions with no writes permitted.
+    """
+    return perms ^ stat.S_IWUSR ^ stat.S_IWGRP ^ stat.S_IWOTH
+
+
+def get_add_write_rule(perms: int) -> int:
+    """
+    Given some existing file permissions, return the same permissions with user and group writes permitted.
+    """
+    return perms | stat.S_IWUSR | stat.S_IWGRP
 
 
 def get_version() -> str:


ZLLentz:
Marking as ready for review, will ask for reviewers after the holiday weekend

ZLLentz:
I'm going to do a tiny bit more cleanup here based on self-review

ZLLentz:
I'm happy with how this behaves now. You'll still be able to undo a release if you'd like to but it won't be possible to do this accidentally now. I think we could discuss a bit about small permutations to this- should we be more protective, and require cross-compiles to unlock the write permissions? Should we mess with the .git directory? Maybe if people have opinions on this in the future we can revisit.

scripts/ioc_deploy.py
```diff
@@ -301,6 +426,143 @@ def make_in(deploy_dir: str, dry_run: bool) -> int:
...
+        )
+    except subprocess.CalledProcessError as exc:
+        return exc.returncode
+    deploy_path = Path(deploy_dir)
+    tracked_paths = [
+        deploy_path / subpath.strip() for subpath in ls_files_output.splitlines()
+    ]
+    build_dir_paths = set()
+
+    def accumulate_subpaths(subpath: Path):
```
tangkong, line 484: Is there a reason this is defined within the scope of `set_permissions` rather than just being a standalone function that takes `build_dir_paths`,  `tracked_paths`, `exclude`, and `tracked_paths` as arguments?  

I like to avoid this sort of closure-type scoping but it's not strictly wrong.  It does save a good amount of typing

scripts/ioc_deploy.py
```diff
@@ -301,6 +426,143 @@ def make_in(deploy_dir: str, dry_run: bool) -> int:
...
+    Apply or remove write protection from a deploy repo.
+
+    You may or may not have permissions to do this to releases created by other users.
+
+    allow_write=True involves adding "w" permissions
+    for the owner and for the group. "w" permissions will never be added for other users.
+    We will also add write permissions to the top-level directory.
+
+    allow_write=False involves removing the "w" permissions
+    from all files, and from directories that contain untracked files.
```
tangkong, line 440: Maybe I'm just slow here, but what's the difference between the files affected when `allow_write=True` and `allow_write=False`?  
* `allow_write=True`  -->  recursively applies to everything inside `deploy_dir` except for ".git"
* `allow_write=False` --> everything tracked by git and their subdirectories recursively, excepting ".git", but then also things not tracked by git?

Is there something missing from the union of "things tracked by git" and "things not tracked by git"?

tangkong (commented):
I'm interested in what happens when two different users try to undo each others write permission settings here.  I want to give this a closer look but I have some questions to start

scripts/ioc_deploy.py
ZLLentz: I'm using the closure scoping here as a lazy way to mutate my set, there's no technical reason or logic flow why it has to stay this way. It could equivalently be a standalone function or pair of functions that create and return sets, or it could be a function I pass a set into to be mutated.

ZLLentz: I'm happy to change it for better code quality

scripts/ioc_deploy.py
tangkong: Maybe I haven't internalized what the directory structure here looks like

ZLLentz: This is a good question that might prompt some edits

I think it's normal to expect that the set of files affected by these two is the same- maybe I should edit so that it is.

The perspective here is something like "there are some directory objects that should never be write-protected because the group wants to be ably to casually recompile on a new arch", but there is no such constraint in the other direction.

- `allow_write=True` -> no constraints, (re-)implemented separately in the simplest way without constraint logic
- `allow_write=False` -> please allow my to be able to log into the rocky9 build host and create e.g. `bin/rhel9_x86_64` even if I've already built/protected once e.g. `bin/rhel7_x86_64`

When implementing the `False` variant with the above goal, I decided that a simple way to do this is to write protect directories that contain files generated from `make`, e.g. those not tracked by git.

Cons of this approach:

- Confusing code
- git safedir thing will pop up when you try to write protect someone else's release

Alternates I considered that maybe you think are better:

- Make it simpler: full write permissions or no write permissions for the entire tree, require using the script to unlock for a new make
- Go for the minimal number of allowed directories: look for folders with OS-names in them, allow only those directories to stay unprotected

I ignored .git because it seemed weird to be messing with it but maybe it's better to mess with it for simplicity's sake

Open to all thoughts and suggestions here

tangkong: Easy agree: I think ignoring .git makes perfect sense

Ok so the difference is that when we set `allow_write=False`, we do not write protect the build artifact directories generated by `make build`

It looks like the code is gathering `build_dir_paths`, then also write-protecting them.  Do we actually want to make these writable?  (are these the `bin/rhel_xx_xx` directories?)

ZLLentz: The `make` generates files and folders such as:
```
bin/rhel7_x86_64/adsIoc
```
Which is the main executable that needs to continue to exist in this example
The intention is to write protect `bin/rhel7_x86_64` without write protecting `bin`:
```
── [drwxrwxr-x]  bin
│   └── [dr-xr-xr-x]  rhel7-x86_64
│       └── [-r-xr-xr-x]  adsIoc
```
So that we can't accidentally remove the folder and delete `adsIoc`, but we can create a parallel folder in `bin` for a new arch.

scripts/ioc_deploy.py
```diff
@@ -301,6 +426,143 @@ def make_in(deploy_dir: str, dry_run: bool) -> int:
...
+        for path in subpath.iterdir():
+            if path.is_symlink():
+                continue
+            elif path.name in exclude:
+                continue
+            elif path.is_dir():
+                accumulate_subpaths(path)
+            elif path.is_file():
+                if path not in tracked_paths:
+                    build_dir_paths.add(str(path.parent))
```
tangkong, line 494: I think I understand this now.  If we come across an untracked file, we add its direct parent to `build_dir_paths`, and that folder gets write protected.  This works for structures like our existing bin folders:
```
bin/   (ignored parent of protected folder)
└── rhel7-x86_64  (protected folder)
    └── ims  (untracked file)

1 directory, 1 file
```

If someone breaks this pattern by accidentally dropping a file in `bin/`, this script breaks right?  (an action which is not explicitly forbidden)

ZLLentz: You're correct, many sorts of manual changes to the build directory will confuse the script and give unexpected outputs. That might be enough of a reason to pick a different methodology.

ZLLentz: fwiw I'm leaning towards simplifying the whole process, write protecting all files and directories with no exceptions, and requiring editing the permissions (using the script) if someone wants to make a new build.

tangkong: I wonder if the simpler (but more brittle) solution is to just add folders like `bin/` to a "folder-exclude" list, but then write protect the contents of these directories.  

The `make` processes may change, but we can probably make some guarantees about what folders the Makefile touches.

tangkong: > write protecting all files and directories with no exceptions, and requiring editing the permissions (using the script) if someone wants to make a new build. 
I can get on board with this.  Even though multi-arch builds are the exception, using the script still provides a path forward

scripts/ioc_deploy.py
```diff
@@ -1,31 +1,39 @@
 #!/usr/bin/python3
 """
 ioc-deploy is a script for building and deploying ioc tags from github.
-It will create a shallow clone of your IOC in the standard release area at the correct path and "make" it.
-If the tag directory already exists, the script will exit.
+
+It will create a shallow clone of your IOC in the standard release area at the
+correct path and "make" it. If the tag directory already exists, the script
+will exit.
```
slactjohnson, line 7: Nit: Maybe this should be updated a bit or an additional paragraph added since now we have an alternate path for just modifying permissions.

scripts/ioc_deploy.py
```diff
@@ -168,9 +219,83 @@ def main(args: CliArgs) -> int:
...
+
+    deploy_dir = get_deploy_info(args).deploy_dir
+
+    if args.allow_write:
+        logger.info(f"Allowing writes to {deploy_dir}")
+    else:
+        logger.info(f"Preventing writes to {deploy_dir}")
+    if not args.auto_confirm:
+        user_text = input("Confirm target? y/n\n")
+        if not user_text.strip().lower().startswith("y"):
```
slactjohnson, line 254: Out of curiosity, why not re-use ```parse_allow_write``` here? The naming is maybe not ideal, but seems to be doing something very similar.

slactjohnson (dismissed):
This looks good to me. Did you test that an IOC release created with the new RO permissions still works as expected? I don't see any reason why it wouldn't, but maybe worth a quick test.

ZLLentz:
> Did you test that an IOC release created with the new RO permissions still works as expected?

I have not done this, but you are right that I need to check a few IOCs to make sure we aren't writing out files to the release directory that then get blocked by the permissions gating.

scripts/ioc_deploy.py
ZLLentz: This is fair, I can update the readme more for the second path including a new example.

scripts/ioc_deploy.py
ZLLentz: @slactjohnson do you have any opinions about this specific topic? Making the permissions handling simpler vs leaving them as-is vs changing how it works for these sorts of edge cases?

scripts/ioc_deploy.py
ZLLentz: You're right, I can re-use `parse_allow_writes` in all of the user confirmation sections. This and the other confirmation sections were written prior to me adding `parse_allow_writes` for the cli extension, there was a worse/different CLI handling in the previous iteration and I never went back to consider revisions. I'll make some changes for this.

ZLLentz: I'll also use this opportunity to consolidate the auto_confirm logic as per Robert's offline suggestion

scripts/ioc_deploy.py
slactjohnson: @ZLLentz I think I'm in favor of keeping things simpler, at least to start. I think that it's possible that we have more multi-arch shenanigans going on in the future than we have in the past as we manage the three body problem of rhel5 <-> rhel7 <-> rocky9, but I think if that becomes painful we can re-visit.

ZLLentz: OK, we are largely in agreement then. I will make the handling maximally simple alongside the other adjustments. The old code for more specific targeting will still exist in git history land if we need to bring it back.

ZLLentz:
The final thing I need to do is to follow Tyler's suggestion of running a real IOC.

I'm going to stand up a simple PV notepad test IOC to try this on. I'll go all the way through iocmanager to verify everything looks normal, then clean up my iocData mess.

If this doesn't work, I'll investigate why and see if any specific directories need to be writable.

ZLLentz:
It seemed to work just fine. The PVs are online in the test subnet e.g. `TST:DEPLOY:STRING` responds to puts and gets and maintains its value with autosave after reboot. The deploy directory is fully write-protected:
```
zlentz@psbuild-rocky9-01:~$ tree -p /cds/group/pcds/epics/ioc/tst/pvNotepad/R0.2.0
/cds/group/pcds/epics/ioc/tst/pvNotepad/R0.2.0
├── [-r--r--r--]  LICENSE.md
├── [-r--r--r--]  Makefile
├── [-r--r--r--]  README.md
├── [dr-xr-sr-x]  build
│   ├── [-r--r--r--]  IOC_APPL_TOP
│   ├── [-r--r--r--]  Makefile
│   ├── [dr-xr-sr-x]  archive
│   │   ├── [-r--r--r--]  ioc-tst-pyqt4.archive
│   │   └── [-r--r--r--]  ioc-tst-try-deploy.archive
│   ├── [dr-xr-sr-x]  autosave
│   │   ├── [-r--r--r--]  ioc-tst-pyqt4.req
│   │   └── [-r--r--r--]  ioc-tst-try-deploy.req
│   └── [dr-xr-sr-x]  iocBoot
│       ├── [-r--r--r--]  Makefile
│       ├── [dr-xr-sr-x]  ioc-tst-pyqt4
│       │   ├── [-r--r--r--]  IOC_APPL_TOP
│       │   ├── [-r--r--r--]  Makefile
│       │   ├── [-r-xr-xr-x]  edm-ioc-tst-pyqt4.cmd
│       │   ├── [-r--r--r--]  envPaths
│       │   ├── [-r--r--r--]  ioc-tst-pyqt4.archive
│       │   ├── [-r--r--r--]  ioc-tst-pyqt4.req
│       │   ├── [-r--r--r--]  ioc-tst-pyqt4.sub-arch
│       │   ├── [-r--r--r--]  ioc-tst-pyqt4.sub-req
│       │   ├── [-r-xr-xr-x]  launchgui-ioc-tst-pyqt4.cmd
│       │   ├── [-r-xr-xr-x]  pydm-ioc-tst-pyqt4.cmd
│       │   └── [-r-xr-xr-x]  st.cmd
│       └── [dr-xr-sr-x]  ioc-tst-try-deploy
│           ├── [-r--r--r--]  IOC_APPL_TOP
│           ├── [-r--r--r--]  Makefile
│           ├── [-r-xr-xr-x]  edm-ioc-tst-try-deploy.cmd
│           ├── [-r--r--r--]  envPaths
│           ├── [-r--r--r--]  ioc-tst-try-deploy.archive
│           ├── [-r--r--r--]  ioc-tst-try-deploy.req
│           ├── [-r--r--r--]  ioc-tst-try-deploy.sub-arch
│           ├── [-r--r--r--]  ioc-tst-try-deploy.sub-req
│           ├── [-r-xr-xr-x]  launchgui-ioc-tst-try-deploy.cmd
│           ├── [-r-xr-xr-x]  pydm-ioc-tst-try-deploy.cmd
│           └── [-r-xr-xr-x]  st.cmd
├── [-r--r--r--]  ioc-tst-pyqt4.cfg
└── [-r--r--r--]  ioc-tst-try-deploy.cfg

6 directories, 34 files
```

ZLLentz:
This is ready pending re-review of my commit storm, or pending if we want to try to test or damage `/cds/group/pcds/epics/ioc/tst/pvNotepad/R0.2.0`. I'm somewhat curious if other users can remove it or use the permissions handlers here to un-protect it, for example, but knowing the answers to that won't necessarily change how I feel about this.

tangkong:
As suspected, I cannot change permissions to this directory:

``` bash
(pcds-5.9.0)roberttk@psbuild-rhel7-01:~/src/engineering_tools/scripts(enh_ioc_deploy_perms +)$ ./ioc-deploy -n ioc-tst-pvNotepad -r R0.2.0 --allow-write True
INFO     ioc-deploy: ioc-deploy: checking inputs
INFO     ioc-deploy: Allowing writes to /reg/g/pcds/epics/ioc/tst/pvNotepad/R0.2.0
Confirm target? yes/true or no/false
yes
ERROR    ioc-deploy: [Errno 1] Operation not permitted: '/reg/g/pcds/epics/ioc/tst/pvNotepad/R0.2.0'
ERROR    ioc-deploy: ioc-deploy errored out
```

ZLLentz:
I'm going to try to sudo chown my test repo to Robert to self-test the new messages I just added.

ZLLentz:
```
zlentz@psbuild-rhel7-01:~$ ./github/engineering_tools/scripts/ioc-deploy --name ioc-tst-pvnotepad -r 0.2.0 --allow-write true
INFO     ioc-deploy: ioc-deploy: checking inputs
INFO     ioc-deploy: Allowing writes to /reg/g/pcds/epics/ioc/tst/pvNotepad/R0.2.0
Confirm target? yes/true or no/false
y
ERROR    ioc-deploy: OSError during chmod: [Errno 1] Operation not permitted: '/reg/g/pcds/epics/ioc/tst/pvNotepad/R0.2.0'
ERROR    ioc-deploy: Please contact file owner roberttk or someone with sudo permissions if you'd like to change the permissions here.
ERROR    ioc-deploy: For example, you might try 'sudo chmod -R ug+w /reg/g/pcds/epics/ioc/tst/pvNotepad/R0.2.0' from a server you have sudo access on.
ERROR    ioc-deploy: ioc-deploy errored out
```

scripts/ioc_deploy.py
```diff
@@ -1,35 +1,62 @@
 #!/usr/bin/python3
 """
 ioc-deploy is a script for building and deploying ioc tags from github.
-It will create a shallow clone of your IOC in the standard release area at the correct path and "make" it.
+
+It has two paths: the normal deploy path, and a second path that adjusts
```
tangkong, line 5:
```suggestion
It has two operation modes: the normal deploy mode, and a second mode that adjusts
```

I think that in a tool that cares so much about the filepath, using "paths" here and elswhere in the docs could be a bit confusing.

scripts/ioc_deploy.py
```diff
@@ -233,30 +426,24 @@ def finalize_name(name: str, github_org: str, ioc_dir: str, verbose: bool) -> st
...
-    if not found_suffix:
+    try:
+        suffix = find_casing_in_dir(dir=str(Path(ioc_dir) / area), name=suffix)
+    except RuntimeError:
         logger.info("This is a new ioc, checking readme for casing")
         # Use suffix from readme but keep area from directory search
-        suffix = split_ioc_name(casing_from_readme(name=name, readme_text=readme_text))[2]
+        suffix = split_ioc_name(casing_from_readme(name=name, readme_text=readme_text))[
+            2
+        ]
```
tangkong, line 445: this formatting is a bit tragic to me.  Was ruff being too inflexible about line lengths?

tangkong (approved):
I'm pretty happy with this (maybe now that I understand it more fully).  I have a bit of phrasing nitpicks, not I'm not too worried.  

I do think that it might make more sense to have `allow-writes` be a subcommand of `ioc-deploy`, such that you'd invoke it like:

```bash
$ ioc-deploy update_perms {RW, RO} -n ioc-tst-thing -r R1.0.0   # runs update perms mode
$ ioc-deploy -n ioc-tst-thing -r R1.0.0  # normal deploy process
```

It does seem like tacking on `--allow-write true` as an optional arg would first deploy the IOC, then adjust its permissions.  

if that's too much work or you're sick of fussing with this, I think it's totally skippable.  This will be useful as it is now

scripts/ioc_deploy.py
ZLLentz: It looks like the line here goes up to 89 chars, I think ruff was being protective of this and choosing the latest spot to split on for some reason.

scripts/ioc_deploy.py
ZLLentz: I'll read the rest of the docs with this in mind and brainstorm some adjustments, but mode is also the wrong word because this is the most technically correct way to describe file permissions (file modes)

ZLLentz:
> I do think that it might make more sense to have allow-writes be a subcommand of ioc-deploy

This is good feedback and I think it's pretty easy to implement

scripts/ioc_deploy.py
tangkong: ruff got rid of flake8's wiggle-room option for line lengths, which is a bit of a bummer

ZLLentz: I think for now I'll add an extra var assignment to get an assemblage that ruff doesn't massacre, in the future we can consider changing how flake8/ruff is configured here

ZLLentz:
argparse is a bit restricting as soon as you want to do anything interesting- it was more annoying than expected to get subcommands to behave as I wanted, but I think I addressed everything in a reasonable or close-to-reasonable way

scripts/ioc_deploy.py
```diff
@@ -700,10 +718,40 @@ def _clone(
...
+    """
+    Small trick to help argparse deal with my optional subcommand.
+
+    This will make argv like this:
+    ioc-deploy -p some_path perms ro
+    be interpretted the same as:
+    ioc-deploy perms ro -p some_path
+
+    Otherwise, the first example here is interpretted as if -p was never passed,
+    which could be confusing.
```
tangkong, line 750: This is truly an extra mile QOL addition, but I can't argue with the results.  I think normally sub-commands just don't work if you don't specify it as the first argument, and I'd have left it at that haha

tangkong (approved):
Yea adding a subparser is always finicky in argparse.  We have some rather complicated boilerplate in other places for it.  (looking at you atef)

No more quips from me.  Well done here 👍

slactjohnson (approved):
LGTM. Nice work!

scripts/ioc_deploy.py
```diff
@@ -718,6 +720,23 @@ def _clone(
     return subprocess.run(cmd, **kwds)
 
 
+def print_help_text_for_readme():
```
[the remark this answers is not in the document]
ZLLentz: You're right, I considered making a standalone script that just stdin -> stdout but I couldn't conceptualize it quickly and didn't feel like increasing my scope.

The idea would be something like
```
ioc-deploy --help | readme-ify
```
I think using help text output in general for the readme is really common even if not everything uses argparse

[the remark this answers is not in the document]
ZLLentz: I'm going to make this an issue

PR #196, opened 2024-09-04
FIX: issues with ioc-deploy casing

ZLLentz:
## Description

Fix an issue where `ioc-deploy` would happily create new folders if you make a casing typo, e.g. `ioc-common-gigecam` will now deploy to `ioc/common/gigECam`.

Github URLs are case-insensitive, e.g. you can clone `ioc-coMMon-GiGeCaM` and it will work just fine.
So, we need to turn to other sources for the correct casing for directory deployment.

The REST API is the most consistent method, but it requires additional authentication.

This PR checks two things in order looking for a match:

- The existing deploy directories
- The README

## Motivation and Context

https://jira.slac.stanford.edu/browse/ECS-6018

## How Has This Been Tested?

Interactively only, works as I expect

Here only

### DIFF
--- a/scripts/ioc_deploy.py
+++ b/scripts/ioc_deploy.py
@@ -209,12 +209,14 @@ def finalize_name(name: str, github_org: str, ioc_dir: str, verbose: bool) -> st
             raise ValueError(
                 f"Error cloning repo, make sure {name} exists in {github_org} and check your permissions!"
             ) from exc
-        try:
-            with open(Path(tmpdir) / name / "README.md", "r") as fd:
-                readme_text = fd.read()
+        readme_text = ""
+        # Search for readme in any casing with any file extension
+        pattern = "".join(f"[{char.lower()}{char.upper()}]" for char in "readme") + "*"
+        for readme_path in (Path(tmpdir) / name).glob(pattern):
+            with open(readme_path, "r") as fd:
+                readme_text += fd.read()
             logger.debug("Successfully read repo readme for backup casing check")
-        except FileNotFoundError:
-            readme_text = ""
+        if not readme_text:
             logger.debug("Unable to read repo readme for backup casing check")
     logger.debug("Checking deploy area for casing")
     # GitHub URLs are case-insensitive, so we need further checks
--- a/scripts/ioc_deploy.py
+++ b/scripts/ioc_deploy.py
@@ -133,7 +133,7 @@ def main(args: CliArgs) -> int:
 
     logger.info("Running ioc-deploy: checking inputs")
     upd_name = finalize_name(
-        name=args.name, github_org=args.github_org, verbose=args.verbose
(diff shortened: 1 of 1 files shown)


scripts/ioc_deploy.py
```diff
@@ -203,9 +209,65 @@ def finalize_name(name: str, github_org: str, verbose: bool) -> str:
...
+
+    found_suffix = False
+    for path in (Path(ioc_dir) / area).iterdir():
+        if path.name.lower() == suffix.lower():
+            suffix = path.name
+            found_suffix = True
+            break
+    if not found_suffix:
+        logger.info("This is a new ioc, checking readme for casing")
+        return casing_from_readme(name=name, readme_text=readme_text)
```
tangkong, line 251: Here if we can't find the suffix, we default to the readme casing, ignoring the area casing we found before.  

Do we want to hold onto the area casing if we find it, in case the readme has some whack non-existent area casing?  Maybe this involves splitting the readme casing output into parts.

tangkong (commented):
I'm on board with this on the whole

scripts/ioc_deploy.py
```diff
@@ -203,9 +209,65 @@ def finalize_name(name: str, github_org: str, verbose: bool) -> str:
             raise ValueError(
                 f"Error cloning repo, make sure {name} exists in {github_org} and check your permissions!"
             ) from exc
+        try:
+            with open(Path(tmpdir) / name / "README.md", "r") as fd:
```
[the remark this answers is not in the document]
ZLLentz: This is fair, I could be more permissive with the kinds of files to check. I originally wanted to avoid answering the question "what do I do when there are two readme files?"

scripts/ioc_deploy.py
ZLLentz: Good catch

scripts/ioc_deploy.py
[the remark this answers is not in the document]
ZLLentz: The answer I'm picking is "include them all in some arbitrary order and do our best"

scripts/ioc_deploy.py
ZLLentz: I did as you described via some more splitting and handling

ZLLentz:
I also ended up adding a check for readmes like this one that have the last part of the suffix in them but not the ioc-area part: https://github.com/pcdshub/ioc-common-ads-ioc/blob/master/README.md

ZLLentz:
Testing for this has been lots of `ioc-deploy --name ioc-CoMmOn-gIgECaM -r R5.0.4` and similar, and it seems to work pretty well.

tangkong (approved):
I like this 👍

slactjohnson (approved):
LGTM! 👍

ZLLentz:
Thank you for the reviews!

PR #199, opened 2024-09-06
ENH: check for github early to avoid long timeouts, decrease usage of git clone

ZLLentz:
## Description

This should be the second-to-last PR from me about this script from the Jira epic.

Main goal: make `ioc-deploy` succeed and fail faster, avoiding slow commands and long timeouts.

Main changes:
- Before a deploy action, `ping` github as a quick connectivity check.
- Check tags, auth, and repo availability all at once using `git ls-remote` one time instead of through multiple `git clone` commands.
- Rearrange logic in `get_deploy_info`, `finalize_name`, and `finalize_tag` to facilitate the above.

Incidental changes:
- Normalize `DeployInfo` to only use `str`, not `None` because it simplified some implementation details
- Rearrange some of the docstrings to match where code has moved
- Allow providing `--path-override` by itself during the deploy routines, which previously would error out. This fell out naturally from the rest of the refactor and I didn't realize it hadn't been supported previously.

## Motivation and Context

Full details at https://jira.slac.stanford.edu/browse/ECS-6096

If you `git clone` from a server like `las-console` or `rix-control`, one of the following will happen:
- If you have psproxy configured, you will be prompted for your password separately on each clone.
- If you do not have psproxy configured, you will time out after 2 minutes.

Both of these cases are avoided by this PR. I wanted to avoid setting timeouts on git clone because the command can take an arbitrarily long amount of time based on the network situation and file sizes. For the same reason, I wanted to avoid calling git clone multiple times if it could be avoided.

For example: cloning the gigECam ioc can often be slow, so I'd like this script to only do it once.

## How Has This Been Tested?

Interactively, extensively

Only in this PR and release notes. The expected results of running `ioc-deploy` remains unchanged, it should just succeed and fail faster.

### DIFF
--- a/scripts/ioc_deploy.py
+++ b/scripts/ioc_deploy.py
@@ -219,6 +219,11 @@ def main_deploy(args: CliArgs) -> int:
 
     Will either return an int return code or raise.
     """
+    logger.info("Checking github connectivity")
+    if not get_github_available(verbose=args.verbose):
+        logger.error("Github is not reachable, please check to make sure you're on a psbuild host.")
+        return ReturnCode.EXCEPTION
+
     deploy_info = get_deploy_info(args)
     deploy_dir = deploy_info.deploy_dir
     pkg_name = deploy_info.pkg_name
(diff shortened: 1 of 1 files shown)


tangkong (approved):
This looks like a straight improvement to me.  This is on the cusp of deserving its own repo / package, this file is quite long..

Whoops about the `--path-override` option.  Should have seen that as a reviewer 😞

slactjohnson (approved):
LGTM!

ZLLentz:
> This is on the cusp of deserving its own repo / package, this file is quite long..

The next time this needs a sizable scope increase, I think the first step is to create an `ioc_deploy` subfolder and make a mini ioc deploy module within engineering_tools, to at least modularize the logic without gaining the weight of formal packaging. If that becomes a bit bigger or we end up adding non-standard-library dependencies then I'd want to tear it out into its own repo.

tangkong:
I wondered if `ghapi` would have saved you some work, but being part of engineering tools makes me value minimal dependencies a bit higher

ZLLentz:
I'm intentionally avoiding `ghapi` sorts of things here anyway (despite the power/convenience) because of the anonymous access api rate limits. I don't want people to have this script error out because someone else used psbuild-rhel7-01 earlier to test their other ghapi script and forgot to authenticate.

ZLLentz:
The other problem with using `ghapi` is that you'd actually need to make a token for non-public repos which is a bit much for what this script is trying to do.

PR #201, opened 2024-09-12
DOC: minor docstring errors from renamed cli args

ZLLentz:
## Description

In one of the ioc-deploy PRs we adjusted the cli arg specifics which made these docstrings incorrect.
This PR fixes a few minor docstrings that I noticed this morning.

## Motivation and Context

Incorrect docs can be confusing

## How Has This Been Tested?

No functional changes to the code

Here only


tangkong (approved):
~Did you re-run the readme-conversion script here? Do we need to?~  Nope these are function docstrings not argparse help text 👍

PR #203, opened 2024-09-27
use daqutils for daqmgr hutches

monarin:
## Description

- Procmgr is being replaced with slurm for some of the hutches. Daqmgr is the new utility for controlling DAQ slurm jobs. The changes here aim at overriding daq related commands such as restartdaq, stopdaq, etc for daqmgr hutches. These commands uses standard python that does not require sourcing prior to calling them. 
- The top part of daq-related scripts checks if the current hutch is daqmgr hutch and overrides the commands with the python modules added here. All the non daqmgr hutches continue to use existing shell scripts.
## Motivation and Context

All slurm jobs launched by daqmgr are stored in slurm database. Daqutils added here is written in python to ease the job of reading of and parsing slurm job statuses. Some of these utilities are slimmed down version of what already exist in daqmgr making it easier to future maintenance work.

Replace procmgr with slurm.
## How Has This Been Tested?

The three commands (restartdaq, stopdaq, and wheredaq) have been tested at tmo and rix.

Similar to the existing requirement, **/engineering_tools/scripts must be part of $PATH for -opr user accounts.

The changes are minimal. None of the preexisting code have been modified apart from the top part of the scripts, which allow overriding daq-related commands.

### DIFF
--- a/scripts/daq_utils.py
+++ b/scripts/daq_utils.py
@@ -26,11 +26,7 @@ def silentremove(filename):
             raise
 
 def call_subprocess(*args):
-    cc = subprocess.run(args, stdout=PIPE, stderr=PIPE)
-    output = None
-    if not cc.returncode:
-        output = str(cc.stdout.strip(), "utf-8")
-    return output
+    return subprocess.check_output(args, stderr=PIPE, universal_newlines=True)
 
 def call_sbatch(cmd, nodelist, scripts_dir):
     sb_script = "#!/bin/bash\n"
--- a/scripts/daq_utils.py
+++ b/scripts/daq_utils.py
@@ -10,13 +10,14 @@ import os, sys
 import getpass
 import time
 
-DAQBATCH_HUTCHES=['tmo', 'rix']
+DAQBATCH_HUTCHES = ["tmo", "rix"]
 LOCALHOST = socket.gethostname()
-SLURM_PARTITION='drpq'
-SLURM_JOBNAME='submit_daq'
-DAQBATCH_SCRIPT='submit_daq.sh'
-DAQBATCH_OUTPUT='slurm_daq.log'
-MAX_RETRIES=10
+SLURM_PARTITION = "drpq"
+SLURM_JOBNAME = "submit_daq"
+DAQBATCH_SCRIPT = "submit_daq.sh"
+DAQBATCH_OUTPUT = "slurm_daq.log"
+MAX_RETRIES = 10
+
 
 def silentremove(filename):
     try:
@@ -25,8 +26,10 @@ def silentremove(filename):
         if e.errno != errno.ENOENT:  # errno.ENOENT = no such file or directory
             raise
 
+
 def call_subprocess(*args):
     return subprocess.check_output(args, stderr=PIPE, universal_newlines=True)
+
 
 def call_sbatch(cmd, nodelist, scripts_dir):
     sb_script = "#!/bin/bash\n"
@@ -42,20 +45,20 @@ def call_sbatch(cmd, nodelist, scripts_dir):
     daqbatch_script = os.path.join(scripts_dir, DAQBATCH_SCRIPT)
     with open(daqbatch_script, "w") as f:
         f.write(sb_script)
-    call_subprocess('sbatch', daqbatch_script)
+    call_subprocess("sbatch", daqbatch_script)
     silentremove(daqbatch_script)
+
 
 class SbatchManager:
     def __init__(self, user):
         self.user = user
 
-
     def get_job_info(self):
         # Use squeue to get JobId, Comment, JobName, Status, and Reasons (Node List)
         format_string = '"%i %k %j %T %R"'
         lines = call_subprocess(
-                    "squeue", "-u", self.user, "-h", "-o", format_string
-                ).splitlines()
+            "squeue", "-u", self.user, "-h", "-o", format_string
+        ).splitlines()
         job_details = {}
         for i, job_info in enumerate(lines):
             cols = job_info.strip('"').split()
@@ -89,33 +92,31 @@ class SbatchManager:
 
 
 class DaqManager:
-    def __init__(self, opts):
-        self.opts = opts
-        self.verbose = False
-        for o, a in self.opts:
-            if o in ('-v', '--verbose'):
-                self.verbose = True
+    def __init__(self, verbose=False):
+        self.verbose = verbose
         self.hutch = call_subprocess("get_info", "--gethutch")
-        self.user = self.hutch+'opr'
+        self.user = self.hutch + "opr"
         self.sbman = SbatchManager(self.user)
-        self.scripts_dir = f'/reg/g/pcds/dist/pds/{self.hutch}/scripts'
+        self.scripts_dir = f"/reg/g/pcds/dist/pds/{self.hutch}/scripts"
         self.set_cnf_file_and_platform()
 
     def isdaqbatch(self, quiet=False):
         if self.hutch in DAQBATCH_HUTCHES:
-            if not quiet: print('true')
+            if not quiet:
+                print("true")
             return True
         else:
-            if not quiet: print('false')
+            if not quiet:
+                print("false")
             return False
 
     def isvaliduser(self):
-        if getpass.getuser().find('opr') > -1:
+        if getpass.getuser() == self.user:
             return True
         return False
 
     def waitfor(self, subcmd):
-        if subcmd == 'stop':
+        if subcmd == "stop":
             daq_host = self.wheredaq(quiet=True)
             if daq_host is not None:
                 if self.isdaqbatch(quiet=True):
@@ -123,52 +124,57 @@ class DaqManager:
                         daq_host = self.wheredaq(quiet=True)
                         if daq_host is None:
                             break
-                        print(f'wait for daqbatch to stop #retry: {i_retry}')
+                        print(f"wait for daqbatch to stop #retry: {i_retry}")
                         time.sleep(1)
                 else:
-                    print(f'the DAQ did not stop properly (DAQ HOST: {daq_host}), exit now and try again or call your POC or the DAQ phone')
+                    print(
+                        f"the DAQ did not stop properly (DAQ HOST: {daq_host}), exit now and try again or call your POC or the DAQ phone"
+                    )
 
     def set_cnf_file_and_platform(self):
-        self.cnf_file = f'{self.hutch}.py'
-        
-        cc = {'platform': None}
+        self.cnf_file = f"{self.hutch}.py"
+
+        cc = {"platform": None}
         configfilename = os.path.join(self.scripts_dir, self.cnf_file)
         try:
-            exec(compile(open(configfilename).read(), configfilename, 'exec'), {}, cc)
-            if type(cc['platform']) == type('') and cc['platform'].isdigit():
-                rv = int(cc['platform'])
+            exec(compile(open(configfilename).read(), configfilename, "exec"), {}, cc)
+            if type(cc["platform"]) == type("") and cc["platform"].isdigit():
+                rv = int(cc["platform"])
         except:
-            print('deduce_platform Error:', sys.exc_info()[1])
+            print("deduce_platform Error:", sys.exc_info()[1])
             raise
         self.platform = rv
 
     def wheredaq(self, quiet=False):
-        """ Locate where the daq is running.
-        
+        """Locate where the daq is running.
+
         For daqbatch, we use slurm to check if the hutch user is running control_gui.
         """
         daq_host = None
         # Use control_gui job name to locate the running host for the daq
         job_details = self.sbman.get_job_info()
-        if 'control_gui' in job_details:
-            daq_host = job_details['control_gui']['nodelist']
-        
+        if "control_gui" in job_details:
+            daq_host = job_details["control_gui"]["nodelist"]
+
         if not quiet:
-            if daq_host is None: 
-                print(f'DAQ is not running in {self.hutch}')
+            if daq_host is None:
+                print(f"DAQ is not running in {self.hutch}")
             else:
-                print(f'DAQ is running on {daq_host}')
+                print(f"DAQ is running on {daq_host}")
         return daq_host
-    
+
     def calldaq(self, subcmd, daq_host=None):
-        prog = 'daqbatch'
-        cmd = f'pushd {self.scripts_dir}'+' > /dev/null;'
-        cmd += f'source {os.path.join(self.scripts_dir, "setup_env.sh")}'+';'
-        cmd += f'WHEREPROG=$(which {prog}); set -x; $WHEREPROG {subcmd} {os.path.join(self.scripts_dir,self.cnf_file)}'+'; { set +x; } 2>/dev/null;'
-        cmd += f'popd'+' > /dev/null;'
-        
-        if subcmd == 'restart':
-            query_daq_host = self.wheredaq() 
+        prog = "daqbatch"
+        cmd = f"pushd {self.scripts_dir}" + " > /dev/null;"
+        cmd += f'source {os.path.join(self.scripts_dir, "setup_env.sh")}' + ";"
+        cmd += (
+            f"WHEREPROG=$(which {prog}); set -x; $WHEREPROG {subcmd} {os.path.join(self.scripts_dir,self.cnf_file)}"
+            + "; { set +x; } 2>/dev/null;"
+        )
+        cmd += f"popd" + " > /dev/null;"
+
+        if subcmd == "restart":
+            query_daq_host = self.wheredaq()
         else:
             query_daq_host = self.wheredaq(quiet=True)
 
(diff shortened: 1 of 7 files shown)


ZLLentz:
I'm going to reformat your description here to restore the "Description" header, approve the workflow run, and add a few people as reviewers

ZLLentz:
For reviews, I'm pinging Silke because she has most of the commits on the previous `restartdaq` script, but she may not have time, so I'm also pinging Vincent because he has the most commits on this repo out of all the CDSOs.

scripts/daq_utils.py
```diff
@@ -0,0 +1,226 @@
...
+                    "state": state,
+                    "nodelist": nodelist,
+                    "logfile": logfile,
+                }
+        return job_details
+
+
+class DaqManager:
+    def __init__(self, opts):
+        self.opts = opts
```
ZLLentz, line 97: I'm concerned about maintainability of scripts that pass CLI options through like this. Someone in the future is going to get confused and need to flip back and forth between this script and the CLI spec to understand what this is supposed to be. It's already kind of hard to follow as-written because there's no clear indication here of what self.opts might be. It'd be cleaner to pass through the different cli options as init args with full names, which also makes this simple to use outside of the CLI.

Something similar to:
```
def __init__(self, daq_host, verbose=False, some_other_option=False):
    self.daq_host = daq_host
    self.verbose = verbose
    self.some_other_option = some_other_option
```

scripts/daq_utils.py
```diff
@@ -0,0 +1,226 @@
...
+    def isdaqbatch(self, quiet=False):
+        if self.hutch in DAQBATCH_HUTCHES:
+            if not quiet: print('true')
+            return True
+        else:
+            if not quiet: print('false')
+            return False
+
+    def isvaliduser(self):
+        if getpass.getuser().find('opr') > -1:
```
ZLLentz, line 117: This will find other users, not just hutch operators, if they happen to have `opr` in their usernames.
There are currently two such usernames.
I think you should instead check for `tmoopr` and `txiopr` explicitly using your `DAQBATCH_HUTCHES` variable.

scripts/daqutils
```diff
@@ -0,0 +1,19 @@
...
+popd () {
+    command popd "$@" > /dev/null
+}
+
+
+# Go to source directory to make python import works
+SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
+pushd $SCRIPT_DIR
+python3 run_daq_utils.py $@
+popd
```
ZLLentz, line 18: There's no need to `pushd` and `popd` inside a bash script. You can `cd` and the calling shell will always stay in the original location. This would only be needed if this script was being sourced.

scripts/run_daq_utils.py
```diff
@@ -0,0 +1,27 @@
...
+help_messages = {
+        'wheredaq': 'Discover what host is running the daq in the current hutch, if any.',
+        'stopdaq': 'Stop the daq in the current hutch.',
+        'restartdaq': 'Verify requirements for running the daq then stop and start it.',
+        'isdaqbatch': "Determine if the current user's hutch uses daqbatch",
+        }
+
+if __name__ == "__main__":
+    commands = ['wheredaq', 'stopdaq', 'restartdaq', 'isdaqbatch']
+    opts, args = getopt.getopt(sys.argv[2:], 'hm:dv', ["help","dss"])
```
ZLLentz, line 14: Consider defaulting to `argparse` instead of `getopt` to get free consistent help text and more readable CLI input parsing.

ZLLentz (commented):
I don't know much about slurm etc. but I tried to give this a review from a code readability/maintainability perspective. Hopefully someone else can verify the functionality.

The CI is giving a lot of errors because the python file doesn't follow pep8. I recommend reformatting with `black` or `ruff` as an easy way to make this consistent.

scripts/daq_utils.py
```diff
@@ -0,0 +1,226 @@
...
+            if daq_host is None: 
+                print(f'DAQ is not running in {self.hutch}')
+            else:
+                print(f'DAQ is running on {daq_host}')
+        return daq_host
+    
+    def calldaq(self, subcmd, daq_host=None):
+        prog = 'daqbatch'
+        cmd = f'pushd {self.scripts_dir}'+' > /dev/null;'
+        cmd += f'source {os.path.join(self.scripts_dir, "setup_env.sh")}'+';'
```
tangkong, line 170: To confirm my understanding: The way we invoke these `daqbatch` commands is by first
* sourcing the `setup_env.sh` script, which activates a `ps-x.x.x` conda environment
* Calling the `daqbatch` command line entrypoint and its subcommands

This environment has access to `psdaq`, so is there a reason we don't just activate this environment once at the beginning of the the script and directly call on `psdaq`'s utilities?  There's a lot of brittle-looking runtime script generation and execution, which might be avoidable

scripts/daq_utils.py
```diff
@@ -0,0 +1,226 @@
...
+                else:
+                    print(f'the DAQ did not stop properly (DAQ HOST: {daq_host}), exit now and try again or call your POC or the DAQ phone')
+
+    def set_cnf_file_and_platform(self):
+        self.cnf_file = f'{self.hutch}.py'
+        
+        cc = {'platform': None}
+        configfilename = os.path.join(self.scripts_dir, self.cnf_file)
+        try:
+            exec(compile(open(configfilename).read(), configfilename, 'exec'), {}, cc)
```
tangkong, line 141: I am generally very uncomfortable with executing external code like this.   At best it obfuscates what is happening (variables can be overwritten, side effects are unknown), and at worst it's a security vulnurability.

Are these configuration settings not something that we can pull in from a more reputable source?

scripts/daq_utils.py
```diff
@@ -0,0 +1,226 @@
...
+        self.calldaq('start', daq_host=daq_host)
+
+    def restart_daqbatch(self, daq_host):
+        st = time.monotonic()
+        self.calldaq('restart', daq_host=daq_host)
+        en = time.monotonic()
+        print(f'took {en-st:.4f}s. for starting the DAQ')
+
+    def restartdaq(self):
+        if not self.isvaliduser():
```
tangkong, line 209: Why is this check not applied to the other daq operations?  Is this the only restricted one?

scripts/run_daq_utils.py
silkenelson: Upvote for argparse. In order to get this rolled out though, could we create an issue to solve in a future and hopefully near update @ZLLentz ?

scripts/daq_utils.py
silkenelson: To my understand psdaq talks to a DAQ that is already running. We are trying to start the DAQ here.

scripts/daq_utils.py
silkenelson: This is just how the DAQ works. We need an editable text file and if you mess if up, you've broken the DAQ. Adding more control will do more harm than help, at the very least in the foreseeable future.  
What I have told Mona to do is to keep a single command that all hutches (LCLS1 and LCLS2 ) can use to start the DAQ., Otherwise we will do _exactly_ what she is doing now, but having to teach everyone a different command and without a way to ensure that they are the users and potentially do more checks that a given process has finished appropriately.

scripts/daq_utils.py
tangkong: This method is used for all the daq commands, as written in this code (start, stop, restart, etc)

scripts/daq_utils.py
silkenelson: the check is only in restartdaq as people who start the DAQ may modify the configuration database and we wanted to make sure that some beamline scientists umask did not make the DAQ unusable for the next person on shift. This is not a problem for the new daq as far as I can see.
We should think about where we need these checks, but to roll this our before SC beam this may need to end up as a item for future review (like argpase versus getopt)

scripts/daq_utils.py
silkenelson: Actually, you create self.user and we basically want to check if a given user is self.user @monarin ?

scripts/run_daq_utils.py
ZLLentz: @silkenelson we can definitely ignore any of the comments per our discretion and promise to do them later, but I caution against incorporating code in our repos that we will have friction maintaining later. It's better to do it cleanly the first time.

If we are to become maintainers of this (as implied by a PR into engineering tools) then we should be at least a little bit picky. If the daq team is going to be the maintainer of this then this PR shouldn't have so much code in it, these scripts should live elsewhere and only be referenced here (or otherwise included in our startup scripts)

I understand of course that this PR is necessary to continue smooth operation of the DAQ given the context. So if you think it's OK feel free to approve and merge.

scripts/daq_utils.py
```diff
@@ -0,0 +1,226 @@
...
+        cc = {'platform': None}
+        configfilename = os.path.join(self.scripts_dir, self.cnf_file)
+        try:
+            exec(compile(open(configfilename).read(), configfilename, 'exec'), {}, cc)
+            if type(cc['platform']) == type('') and cc['platform'].isdigit():
+                rv = int(cc['platform'])
+        except:
+            print('deduce_platform Error:', sys.exc_info()[1])
+            raise
+        self.platform = rv
```
ZLLentz, line 147: Small issues in the code here that I only noticed after reading other comments:
- `rv` is not guaranteed to be set, so it's possible to get us into a state where `self.platform = rv` raises a confusing `NameError`
- This is not the normal way to do type checking in Python, and this type check isn't necessary
- bare excepts are not great because they catch things like `KeyboardInterrupt`

I think instead we should lean on the built-in type checking from `int` and fail out into the decide platform error print if there's an issue.

```suggestion
            self.platform = int(cc['platform'])
        except Exception:
            print('deduce_platform Error:', sys.exc_info()[1])
            raise
```

scripts/daq_utils.py
ZLLentz: It does feel at least a little bit silly to exec an entire file to collect a single integer configuration parameter. If that's what's required by the DAQ right now some less hazardous specifications should be considered.

monarin:
HI all,  sorry for my delayed response. I really appreciate a lot of comments and feedbacks here! Sorry if I can't reply directly to all these suggestions now, I hope to be sitting down and working with both chris and silke next week so we can work through these. Hopefully the next version will be more ready.

scripts/daq_utils.py
silkenelson: I will now own up to my ignorance - I think this is due to a misunderstanding that Mona had when she read the original restartdaq that works w/ the procMgr setup. For the DAQ we need a 'platform' which is coded up in the *cnf(.py) file that you want to run. I had also used 'platform' to determine which *cnf file to run in the LCLS1 setup: e.g. CXI had two DAQs that used different platforms and we needed to pick if to start cxi_0.cnf or cxi_1.cnf.  So we really do not need to read the platform back from the cnf file, we only need to come up with a way of picking the right cnf file. The use case here is e.g. RIXS where they will use one cnf.py file for chemRIXS and another one for qRIXS.
This is the part where we have to sit together and make sure we are doing the right thing. A combination of Chris, Mona, myself and Patrick is likely optimal.

ZLLentz:
I should have mentioned this last week, but the core intent and implementation here (the main points in the PR description, such as switching to slurm) were good and didn't need any scrutiny. Aside from some questions for understanding, my comments are largely targeting maintainability (putting all the cli parsing in one place, avoiding confusing error tracebacks, etc.).

I hope I didn't come across as too critical and I'm sorry if I did.

scripts/daq_utils.py
monarin: Resolved: check user with DAQBATCH_HUTCHES + 'opr'

scripts/run_daq_utils.py
monarin: Fixed. The use of argparse is added in the main file. This also resolves the issues with arguments being parsed at different places in the code.

scripts/daqutils
monarin: Changed this to cd.

scripts/daq_utils.py
```diff
@@ -0,0 +1,226 @@
...
+            raise
+
+def call_subprocess(*args):
+    cc = subprocess.run(args, stdout=PIPE, stderr=PIPE)
+    output = None
+    if not cc.returncode:
+        output = str(cc.stdout.strip(), "utf-8")
+    return output
+
+def call_sbatch(cmd, nodelist, scripts_dir):
```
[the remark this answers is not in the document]
monarin: We can but this is short enough to do it here without creating one.

scripts/daqutils
monarin: Fixed.

scripts/daq_utils.py
```diff
@@ -0,0 +1,214 @@
...
+
+def call_subprocess(*args):
+    # FIXME: potentialy wanted to replace with 
+    # return subprocess.check_output(args, stderr=PIPE, universal_newlines=True)
+    # current block: return values are not in expected format. 
+    cc = subprocess.run(args, stdout=PIPE, stderr=PIPE)
+    output = None
+    if not cc.returncode:
+        output = str(cc.stdout.strip(), "utf-8")
+    return output
```
ZLLentz, line 38: If this can't be replaced with a version that raises on exception, the next best thing is to check for "None" wherever this is used. The important thing I was trying to get at was that there were places elsewhere that assume the result is a `str` and then the user could end up making really bad bug reports like:

`AttributeError: 'NoneType' object has no attribute 'splitlines'`

And then you or me needs to dig into this and figure out what the actual bug is. But if this raises a `CalledProcessError` instead, you get an error message from the script you're calling, which gives us a head start on debugging and can even help the user fix their own problems.

To keep the format the same I would suggest adding `check=True` as an argument to make this raise the helpful exception message.

scripts/daq_utils.py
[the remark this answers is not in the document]
ZLLentz: Ok, makes sense if you found this simpler to implement. I think I'd would have found it simpler to implement the other way but both work.

scripts/daq_utils.py
```diff
@@ -0,0 +1,214 @@
...
+            return True
+        else:
+            if not quiet:
+                print("false")
+            return False
+
+    def isvaliduser(self):
+        if getpass.getuser() == self.user:
+            return True
+        return False
```
ZLLentz, line 123: Nitpick, can ignore
```suggestion
        return getpass.getuser() == self.user:
```

scripts/daqutils
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+
+
+# Go to source directory to make python import works
+SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
+cd $SCRIPT_DIR
```
ZLLentz, line 6: Another tiny nitpick, can ignore
```suggestion
cd -- "$( dirname -- "${BASH_SOURCE[0]}" )"
```

ZLLentz (approved):
Mostly nitpicks left, leaving my approval stamp
I'm still worried for future us getting weird NoneType bug reports but I understand we want to get this in

scripts/daqutils
monarin: This is a tricky one and sorry for my ignorance (I basically copied this from the original restartdaq script). I think it tries to use either the directory of this source file (not sure about this, but maybe users can run it from somewhere else?) or the current directory. It also hushes the output. 
Now thinking about it, I'd probably want to expand it but given that it works in the previous restartdaq, we probably can keep it this way.

silkenelson:
For the old setup, I wanted to cd into the same directory so that the temporary files such as pid.txt would always end up in the same space.

scripts/daq_utils.py
monarin: We determined that restartdaq doesn't need the platform parameter. The function was removed.

ZLLentz:
I think this is good to merge if someone tries it out and confirms it does what they need

scripts/daq_utils.py
monarin: The output without automatic conversion works. This should untrap all the subprocess errors now.

monarin:
We decided to rename the slurm-for-daq manager from daqbatch to daqmgr. This is to avoid technical details in the name. We needed to complete this name change while it's still early enough. Unfortunately, this happened to be in the middle of this PR and I had to complete the work here too to verify the test at tmo and rix.

monarin:
I'm looking at the pre-commit errors. Can anyone advise how to run pre-commit test interactively?

ZLLentz:
First, please ignore any pre-commit warnings from parts you did not edit. We're in the process of cleaning some things up in #204.

Some of the pre-commit jobs will fix the formatting for you, others will just complain.

To run pre-commit interactively:
```
source pcds_conda  # /cds/group/pcds/pyps/conda/pcds_conda if not on PATH
pre-commit install  # optional, makes pre-commit run before every commit for this repo
pre-commit run --files <filename>  # make pre-commit run on a specific file
```

ZLLentz:
(for example, please ignore shellcheck for now- it's not on you to edit the entire restartdaq script)

ZLLentz (approved):
To my eye, the script (still) looks ready to merge!

I would really appreciate it if a CDSO from TMO or RIX could give it a go, but we can also merge first and return later if Tong, Basil, Jyoti, or Patrick have input based on usage.

PR #204, opened 2024-09-29
Shellcheck and shfmt on bash scripts

KaushikMalapati:
## Description

I accidentally changed the branch for the old pr #184  which automatically closed it, so I am making a new one.
I added a pre-commit job for shfmt which will automatically format bash scripts and went through all existing shellcheck errors/warnings that the existing job raises. I am disabling some warnings globally in .shellcheckrc and on a line-by-line basis in a few files. 

## Motivation and Context

https://github.com/pcdshub/engineering_tools/issues/182
To standardize style and fix existing shellcheck issues so that future contributors only see pre-commit/workflow errors related to their changes instead of what was already there.

## How Has This Been Tested?

Interactively

N/A

### DIFF
--- a/scripts/archive-status
+++ b/scripts/archive-status
@@ -12,7 +12,7 @@ if [[ -z $1 ]]; then
     usage
     exit 1
 fi
-    
+
 if [[ ($1 == "--help") || ($1 == "-h") ]]; then
     usage
     exit 0
@@ -20,5 +20,5 @@ fi
 
 result=$(curl -s "http://pscaa01.slac.stanford.edu:17665/mgmt/bpl/getPVStatus?pv=$1")
 echo
-echo $result | cut -d'{' -f2 | cut -d'}' -f1 | tr ',' '\n'
+echo "$result" | cut -d'{' -f2 | cut -d'}' -f1 | tr ',' '\n'
 echo
--- a/scripts/archive-status
+++ b/scripts/archive-status
@@ -1,7 +1,7 @@
 #!/bin/bash
 usage()
 {
-cat << EOF
+    cat << EOF
 usage: $0 <PV>
 
 Check on the status of a PV in the archiver, e.g. $0 GDET:FEE1:241:ENRC
--- a/scripts/configdb_readxtc
+++ b/scripts/configdb_readxtc
@@ -18,7 +18,7 @@ if [[ ($1 == "--help") || ($1 == "-h") ]]; then
 fi
 
 USER=$(whoami)
-ARGSTR=''
+export ARGSTR=''
 
 while getopts "u:e:" OPTION
 do
--- a/scripts/configdb_readxtc
+++ b/scripts/configdb_readxtc
@@ -1,7 +1,7 @@
 #! /bin/bash
 usage()
 {
-cat << EOF
+    cat << EOF
 usage: $0 options
 
 We will run configdb_readxtc
@@ -13,31 +13,31 @@ EOF
 }
 
 if [[ ($1 == "--help") || ($1 == "-h") ]]; then
-	usage
-	exit 0
+    usage
+    exit 0
 fi
 
 USER=$(whoami)
 export ARGSTR=''
 
-while getopts "u:e:" OPTION
-do
+while getopts "u:e:" OPTION; do
     case $OPTION in
-	u)
-	    USER=$OPTARG
-	    ;;
-	e)
-	    EXP=$OPTARG
-	    ;;
-	?)
-	    usage
-	    exit
-	    ;;
-	esac
+        u)
+            USER=$OPTARG
+            ;;
+        e)
+            EXP=$OPTARG
+            ;;
+        ?)
+            usage
+            exit
+            ;;
+    esac
 done
 
 if [[ $USER =~ "opr" ]]; then
-    printf "Please enter user name (cannot run as xppopr): \n"; read -r USER
+    printf "Please enter user name (cannot run as xppopr): \n"
+    read -r USER
 fi
 
 : "${EXP:=$(get_curr_exp)}"
@@ -61,10 +61,10 @@ if [ $ret_code == 0 ]; then
 else
     dir="/reg/g/pcds/dist/pds/xpp/current/build/pdsapp/bin/x86_64-rhel7"
     if eval find $dir/configdb_readxtc; then
-	EXE='/reg/g/pcds/dist/pds/'$HUTCH'/current/build/pdsapp/bin/x86_64-rhel7/configdb_readxtc'
+        EXE='/reg/g/pcds/dist/pds/'$HUTCH'/current/build/pdsapp/bin/x86_64-rhel7/configdb_readxtc'
     else
-	echo 'could not find configdb_readxtc executable'
-	exit
+        echo 'could not find configdb_readxtc executable'
+        exit
     fi
 fi
 
@@ -73,7 +73,7 @@ fi
 #
 if [[ $HOSTNAME =~ "psana" ]]; then
     $EXE /reg/d/psdm/"$HUTCH"/"$EXP"/xtc
-elif  [[ $HOSTNAME =~ "psusr" ]]; then
+elif [[ $HOSTNAME =~ "psusr" ]]; then
     ssh -Y "$USER"@psana "$EXE /reg/d/psdm/$HUTCH/$EXP/xtc"
 else
     echo "$EXE /reg/d/psdm/$HUTCH/$EXP/xtc"
--- a/scripts/configdb_readxtc
+++ b/scripts/configdb_readxtc
@@ -8,7 +8,7 @@ We will run configdb_readxtc
 
 OPTIONS:
 -u user (needs to be able to log into the psananeh/feh)
--e expnumber 
+-e expnumber
 EOF
 }
 
@@ -17,7 +17,7 @@ if [[ ($1 == "--help") || ($1 == "-h") ]]; then
 	exit 0
 fi
 
-USER=`whoami`
+USER=$(whoami)
 ARGSTR=''
 
 while getopts "u:e:" OPTION
@@ -37,10 +37,10 @@ do
 done
 
 if [[ $USER =~ "opr" ]]; then
-    printf "Please enter user name (cannot run as xppopr): \n"; read USER
+    printf "Please enter user name (cannot run as xppopr): \n"; read -r USER
 fi
 
-: ${EXP:=`get_curr_exp`}
+: "${EXP:=$(get_curr_exp)}"
 HUTCH=${EXP:0:3}
 
 # check if experiment is passed, then use that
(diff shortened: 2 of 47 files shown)


ZLLentz (commented):
Partial review - made it up through `checkcnf` and found most things look very clean!
I'll resume my review either later or tomorrow depending on how things go

tangkong:
If I understand correctly, this is a re-creation of the previous shellcheck PR we spent some time reviewing before right?  Or were there other changes made in addition?  If I recall correctly, our main issue was that this touched a lot of files and we don't have a great way of regression testing the changes made here (short of manually testing the scripts). 

I still think this looks largely un-problematic, though I might feel better if we approached this in pieces.  Or maybe we just rip off the bandaid and make sure we're ready to hotfix any unexpected breaks

ZLLentz:
The main differences I see between this and previous are the addition of the shfmt pre-commit and applying fixes to files added since the last PR. I'm going to continue reviewing.

scripts/serverStat
```diff
@@ -118,56 +120,59 @@ fi
...
+        NEVENT=${DEV:5}        # grab the index (e.g. 2 if $DEV is event2)
+        SEDNUM=$((NEVENT + 1)) # adjust the number to correctly index the list of dssnodes with sed
         NAME=$(echo "$DSSNODES" | sed -n ${SEDNUM}p)
         if [[ -z $NAME ]]; then
             echo "dss node not found for $DEV, exiting..." >&2
             exit 1
         fi
     else
-        NDEV=$(grep "$DEV" "$CNF" | grep -v '#' | grep host | wc -l)
+        NDEV=$(grep "$DEV" "$CNF" | grep -v '#' | grep -c host)
```
ZLLentz, line 168: TIL `grep -c`

scripts/archive-status
```diff
@@ -12,13 +12,13 @@ if [[ -z $1 ]]; then
...
     exit 1
 fi
-    
+
 if [[ ($1 == "--help") || ($1 == "-h") ]]; then
     usage
     exit 0
 fi
 
 result=$(curl -s "http://pscaa01.slac.stanford.edu:17665/mgmt/bpl/getPVStatus?pv=$1")
```
tangkong, line 21: For followup: these urls will be deprecated soon.  We should be using `https://pswww.slac.stanford.edu/archiveviewer/retrieval/`

scripts/set_gem_timing
```diff
@@ -8,54 +8,54 @@ IOC_PATH_SC="ioc/fee/GasDetDAQ/R4.0.22-NOTS"
...
-  exit 1
+    echo "Error. Must pass strictly one argument, at least one additional arg: $2, please run again with exclusively [NC or SC] as argument"
+    exit 1
 fi
 
 #get caget and caput
 source /cds/group/pcds/setup/epicsenv-cur.sh #Note: (josh) we should only source the actual executables we are using, this is heavy handed. shellcheck does not love this; however if this path doesnt exist we have bigger problems
 #get imgr
-IMGR_CMD_PATH="/reg/g/pcds/engineering_tools/latest-released/scripts/imgr"
+IMGR_CMD_PATH="/cds/group/pcds/engineering_tools/latest-released/scripts/imgr"
```
ZLLentz, line 31: This is kind of bad behavior that should be addressed in a separate PR. I'm accumulating issues for the repo as I am ending up reading most of the scripts to double-check the reformatting here.

This should use the imgr script from the same release, not have a hard-coded link to whatever the latest release is.

ZLLentz (commented):
The only issue I found with the reformat here is the duplication of the python script
This is still super scary to merge but maybe it's time to just do it

I'm making issues for the other things I found

scripts/configdb_readxtc
```diff
@@ -1,46 +1,46 @@
...
-	usage
-	exit 0
+    usage
+    exit 0
 fi
 
-USER=`whoami`
-ARGSTR=''
+USER=$(whoami)
+export ARGSTR=''
```
tangkong, line 21: why do we export this variable and not the other?  This also assumes we don't source this, otherwise we get `ARGSTR` in our environment variables right?

scripts/new_bare_repo
tangkong, line 1: Is this also something we should deprecate in favor of our more modern ioc deploy processes?

scripts/pcds_conda
```diff
@@ -30,72 +30,72 @@ CONDA_ROOT="${PYPS_ROOT}/conda"
...
+    fi
+    # Figure out where the base environment is
+    # Base environment should be updated with every new python version
+    # But in practice this can be skipped
+    if [[ ${SERIES} == "pcds" ]]; then
+        if version_ge "${VER_NUM}" "${FIRST_PCDS_PY39}"; then
+            BASE_ENV="${CONDA_ROOT}/py39"
+            PYTHON_VER="3.9"
+        elif version_ge "${VER_NUM}" "${FIRST_PCDS_PY38}"; then
+            BASE_ENV="${CONDA_ROOT}/py36"
```
tangkong, line 89: Should this be "py38"?  If we're using `PYTHON_VER="3.8"`?

scripts/set_gem_timing
```diff
@@ -8,54 +8,54 @@ IOC_PATH_SC="ioc/fee/GasDetDAQ/R4.0.22-NOTS"
...
 
 # check current operational state
 OP_MODE=$(caget KFE:CAM:TPR:02:MODE | awk '{print $2}')
 SET_MODE_CMD="caput KFE:CAM:TPR:02:MODE $1"
 
 echo "Current OP_MODE is: $OP_MODE"
-if [[ $OP_MODE = "$1" ]]; then
-  echo "Warning. Already in requested mode: $OP_MODE" #Note: (josh) we are opting to consider this not error worthy
+if [[ $OP_MODE == "$1" ]]; then
```
tangkong, line 38: Interesting that I've seen recommendations that `=` be used for other posix shell compatibility, but since this is explicitly a bash script I suppose this is fine/equivalent

scripts/startami
```diff
@@ -61,46 +60,45 @@ if [[ -z $CONFIG ]]; then
...
 
-DAQHOST=`wheredaq`
-ami_base_path=`grep ami_base_path /reg/g/pcds/dist/pds/$HUTCH/scripts/$CONFIG | grep -v '#' | grep -v 'ami_base_path+'  | awk 'BEGIN { FS = "=" }; { print $2}' | sed s/\'//g`
-ami_path=$ami_base_path`grep ami_path /reg/g/pcds/dist/pds/$HUTCH/scripts/$CONFIG | grep -v 'ami_path+' | grep -v '#' | awk 'BEGIN { FS = "= " }; { print $2}' | sed s/ami_base_path+// | sed s/\'//g`
+proxy_cds=$(/cds/group/pcds/dist/pds/"$HUTCH"/current/tools/procmgr/procmgr status /cds/group/pcds/dist/pds/"$HUTCH"/scripts/"$CONFIG" | grep ami_proxy | awk '{print $1}' | sed s/'-fez'/''/g)
 
-proxy_cds=`/reg/g/pcds/dist/pds/$HUTCH/current/tools/procmgr/procmgr status /reg/g/pcds/dist/pds/$HUTCH/scripts/$CONFIG | grep ami_proxy | awk {'print $1'} | sed s/'-fez'/''/g`
+amicmd=$(grep ami_client /cds/group/pcds/dist/pds/"$HUTCH"/scripts/"$CONFIG" | grep -v '#' | awk 'BEGIN { FS = ":" }; { print $4}' | sed s/ami_path//g | sed s/\'+proxy_cds/"$proxy_cds"/g | sed s:\'+expname:"$EXPNAME"/:g | sed s/+\'//g | sed s/\'\}\)//g)
```
tangkong, line 79: I would love comments explaining what is going on here. 😵   I've not yet conquered `sed`  (don't feel obligated to do this in this PR)

scripts/pcds_conda
ZLLentz: There is no py38 folder in the conda root

tangkong: So there's just py38 environments in the py36 folder? ha

scripts/startami
KaushikMalapati: If the script argument to sed starts with a s, it means you are defining a replacement pattern to apply to the input which can be a file(s) or in this case, piped. The slashes are delimiters - between the first and second slash is the string to search for in the input - for the proxy_cds line, this is '-fez'. Between the second and third slashes is the empty string, which is what to replace the first string with. The g after the third slash means global, so we will replace every instance of '-fez' with ''. Essentially, we are removing all -fez from the output of the awk behind it. The delimited does not have to be a slash - whatever the character after the s is will be used as the delimiter, so it is convenient to choose a delimiter that isn't used in the find and replace strings so you don't have to escape. 

In the amicmd line, there are several sed commands used. `sed s/ami_path//g` replaces every instance of "ami_path" with an empty string. `sed s:\'+expname:"$EXPNAME"/:g` uses a colon as a delimiter so the slash in the replacement string does not have to be escaped - here the literal string "'+expname" is being replaced with the value of the variable $EXPNAME - the single quote is escaped so bash does not look for a closing single quote. The double quotes around $EXPNAME are not part of the replacement string but are there to prevent globbing (https://www.shellcheck.net/wiki/SC2086). 

Not sure if I explained it well, but basically there are bunch of find and replace operations being performed on piped command output.

scripts/set_gem_timing
KaushikMalapati: Using double brackets it also not posix compliant. I do not think anything used at slac would have issues with either any features like this.

scripts/configdb_readxtc
KaushikMalapati: I was too lazy to check if the child process this spawned needed the variables to be set or the empty string specifically so left it in case it was used, and exported it to get rid of the shellcheck warning. Searching through https://github.com/lcls-daq/pdsapp, I don't see ARGSTR used anywhere so it probably should be removed.

rc/bashrc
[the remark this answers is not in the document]
ZLLentz: I think we should, or we could at least include instructions in here for updating to the full shared dotfiles. This is a reasonable follow-up issue.

scripts/new_bare_repo
ZLLentz: This should be entirely removed I think (separate PR), as should anything else that's built for working with afs

scripts/configdb_readxtc
KaushikMalapati: I went through the exports I added and removed most of them. In the cases where I think the variables are completely unused, I deleted them. In others I excused the warning. Specifically,

- Removing NOTRUNNING from wherepsana
- removing STOP_AMI from startami
- Removing test and and if block in restartdaq that does nothing
- Removing the export and disabling warnings on STARTTIME and ENDTIME in motorinfo since a comment suggests they may be used in the future
- Removing ARGSTR from configdb_readxtc
- Removing the export and disabling the warning on S3DF=1 in makepeds. I can't tell what BINPATH resolves to from reading the script so I didn't remove the variable entirely

ZLLentz:
I think this is good if anyone is brave enough to merge it

tangkong:
I'm scared tbh.  After we merge, should we tag/update asap to get people using these scripts?   This will be the root of many a merge conflict I'm sure

silkenelson:
I have a pull request to add needed functionality to makepeds which failed pre-commit which will have to resolved and I don't even have the time to fix the current problems. 
Would there be a chance to not push changes to 56 files at the same time, in particular during operations? We need to make operational needed changes and I'm also concerned this will be messy.

ZLLentz:
Silke, this repo is not gating merges on pre-commit failures, they are currently optional, so for your PR you should proceed as normal given the time investment possible. The only merge requirement here is "one passing review".

For this PR I suspect we want to at least test the scripts that have gotten more than surface-level (whitespace) edits. Maybe the next step for this one is to collect the scripts that need to be re-tested.

KaushikMalapati:
I think these scripts should be retested
- camViewer
- ioctool
- ipmConfigEpics
- makepeds
- makepeds_psana
- motorInfo
- pcds_conda
- restartdaq
- serverStat
- startami
- wherepsana

ZLLentz:
I skimmed the PR again and I think I agree with those choices, some may not necessarily need it but it's better to be thorough.
The only questions are then:

- Who should test these? Maybe we can split up the work a bit?
- How should each be tested?
- Do we need to split up the PR to get in the simple ones first, or is that not worth the effort?

tangkong:
I'm in favor of splitting these up, which to me doesn't sound like too much git-fu but I realize it could be annoying. 

I can't say that I'd be very useful in testing these scripts, but I do dream of a day when the test procedure is a little more robust than "run the script and see if it breaks".  I assume there's been no real testing procedure in place?  Is there even a way to test these without building mock services for all the things these scripts touch?

PR #209, opened 2024-10-02
add xcs usrdio to ipmConfigEpics

patoppermann:
add xcs usrdio to ipmConfigEpics
## Description
xcs usrdio has to be add to ipmConfigEpics  to it can be used from the hutch engineering tools.


patoppermann:
Tested and works.
image

ZLLentz (approved):
Works when I run it too
We can probably ignore the pre-commit and let #204 fix it

Next time best practice is to work on a branch instead of master

PR #218, opened 2024-10-31
BUG: fix second ami client start

vespos:
## Description
startami fails when trying to get a second client

## Motivation and Context
Shell check forcing some changes that do not work. Revert and ignore shell check.

## How Has This Been Tested?
xpp-control as xppopr

N/A

### DIFF
--- a/scripts/startami
+++ b/scripts/startami
@@ -107,4 +107,5 @@ else
 fi
 
 echo "$ami_path""$amicmd"
-exec "$ami_path""$amicmd"&
+# shellcheck disable=SC2086
+$ami_path$amicmd &
--- a/scripts/startami
+++ b/scripts/startami
@@ -107,5 +107,4 @@ else
 fi
 
 echo "$ami_path""$amicmd"
-# shellcheck disable=SC2086
-$ami_path$amicmd &
+eval "$ami_path$amicmd &"


vespos:
@ZLLentz shellcheck seems to be an issue here. I feel we have only recently been using it, do you know why it is suggesting something that does not work here?

gadorlhiac (approved):
I am okay to approve this now since this is broken for every hutch since they all run off latest currently. We could refactor this in a subsequent PR?

The issue specifically is that the `amicmd` is parsed to hold a number of arguments:
```bash
> echo $amicmd
/online_ami -s daq-xpp-mon01 -f ~xppopr/experiments// -C mono,jet
```

I believe we could refactor this to separate out those components and bring it back into compliance with the spellcheck rules. Something like this would also seem to work and pass spellcheck without having to do any specific parsing: `exec $(eval "$ami_path$amicmd")&`

Apologies for not realizing this on the last PR! I tested only the first client which exits early before this command is run.

ZLLentz:
I'll look into this weird command splitting thing in depth, if this works as-is merge it and proceed forward

ZLLentz (approved):

ZLLentz:
Linking PR to slack thread so I can find slack thread later: https://slac.slack.com/archives/C2PNS4Q74/p1730337117129199

ZLLentz (approved):

gadorlhiac (approved):

Issue #220, opened 2024-11-15
BUG: epicsArchChecker fails to detect if the last alias is missing a pv


KaushikMalapati:
## Expected Behavior

epicsArchChecker should print messages telling the user that there are aliases with no pv if there are any anywhere in the arch file.

## Current Behavior

If there are mutliple aliases and the last alias has no pv, it will not cause a warning to print even though it should. All but the last alias will cause prints if they don't have correspoding pvs.

## Possible Solution

The problem is probably somewhere in the for loop in read_file. I think it might be because we only add alias to the extra keys list if we find another key before finding a line starting with an alnum char, so it's impossible for the last key alias to reach that case
https://github.com/pcdshub/engineering_tools/blob/6d210411ef86459a057e8d1973f21346faf8493b/scripts/epicsArchChecker#L89

## Steps to Reproduce (for bugs)

1. Use this as the arch file

```
*Port_0_MCP_Front
TMO:MPOD:01:M0:C0:VoltageMeasure ca

*Port_1_MCP_Front
#TMO:MPOD:01:M0:C1:VoltageMeasure ca

#test
```

3. Run epicsArchChecker on it 
4. Notice that there is no error even though the last alias clearly has no pv.

## Context

Found it by accident while working on https://github.com/pcdshub/engineering_tools/pull/219. Hasn't affected me, but it would be unfortunate if the script failed to catch an error because it happened to be at the end of the file and it caused daq problems. It would be hard to diagnose because this script would incorrectly tell people that the arch file is not the problem. 

## Your Environment

n/a

PR #221, opened 2024-11-16
Fixing epicsarch bug on last alias not having a pv

KaushikMalapati:
## Description

Probably should have included this in #219
I'm just checking if key is not empty after the for loop in read_file. If it is, that means we did not encounter a pv after the last alias, and it should be in extraKeys, but was not because the for loop only does so when it encounters the next key (a line that starts with * when key is not empty), which does not work on the last alias.

## Motivation and Context

#220 

## How Has This Been Tested?

Interactively with variations of this arch file
```
*Port_0_MCP_Front
TMO:MPOD:01:M0:C0:VoltageMeasure ca

*Port_1_MCP_Front
#TMO:MPOD:01:M0:C1:VoltageMeasure ca

#test
```

n/a

### DIFF
--- a/scripts/epicsArchChecker
+++ b/scripts/epicsArchChecker
@@ -122,6 +122,8 @@ def read_file(filename):
                     else:
                         entries.append((key, pv, filename, lineNum))
                         key = ''
+            if key != '':
+                extraKeys.append((key, filename, keyline))
     except FileNotFoundError as errorDetail:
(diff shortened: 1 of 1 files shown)


tangkong:
For the record (hah) I think it's more apt as two separate PRs, they're accomplishing two different tasks 👍

ZLLentz (approved):

PR #223, opened 2024-11-20
Using epics7 env for rocky9 compatibility

KaushikMalapati:
## Description

motor-expert-screen was using epicsenv-3.14.12.sh, but the oldest version of epics base available on rocky9 is 7.0.2, so motor-expert-screen will use that environment from now on

I also had to update how :PN motors are opened, because originally they executed a file that tried to activate the epics 3.14 environment. Now I am directly opening the screen with edm and I added the path to EDMDATAFILES

## Motivation and Context

https://jira.slac.stanford.edu/browse/ECS-6842

## How Has This Been Tested?

Interactively

n/a

### DIFF
--- a/scripts/motor-expert-screen
+++ b/scripts/motor-expert-screen
@@ -23,7 +23,7 @@ PREFIX=`echo $1 | cut -d . -f 1`
 source /reg/g/pcds/setup/epicsenv-7.0.3.1-2.0.sh
 export PCDS_EDMS=/reg/g/pcds/package/epics/3.14/screens/edm
 #export EDMDATAFILES=.:${PCDS_EDMS}/xps8:${PCDS_EDMS}/ims  # normal EDMDATAFILES line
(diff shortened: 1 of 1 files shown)


ZLLentz:
I agree with this conceptually and I agree with the implementation
Let me double-check a few screen loads before/after on a hutch machine to make sure there aren't any obvious regressions

ZLLentz:
This PR absolutely enables this script to be run from rocky9 hosts! Very good.

Here are the PV bases I'm planning to compare:

IMS: XPP:USR:MMS:01
PP: XPP:SB2:MMS:29
Newport: XPP:USR:MMN:01
AeroTech: XCS:SND:T1:Y1
SmarAct: XPP:MCS2:01:m1
Beckhoff (old): MFX:LJH:JET:X
Beckhoff (new): IM3L0:PPM:MMS

So far, the IMS was the exactly the same but the PP (old IMS) picked a slightly different screen (top is with the PR, bottom is before this PR):
image

I'll continue through the rest, it shouldn't take too long

ZLLentz:
Everything was the same except for these super old pulse picker screens which are now using a less readable variant

ZLLentz:
I think if for a bunch of PVs and on the hutch operator machines (-daq, -control) we get the same results, then everything should be good.

We want to enable this on rocky9 but not at the expense of ongoing experiments who may rely on this script.

ZLLentz:
I think you've done more than enough testing!
I had forgotten that the typhos screens don't open on rhel5. Maybe this is OK though (and certainly it predates this PR)

ZLLentz (approved):

PR #225, opened 2024-11-20
grep_more_ioc fix pandas silent downcasting spam

### LINKED ISSUE #224: Pandas 3.0 future warning on downcasting spam
## Expected Behavior
Turns out pandas was doing some silent downcasting when you call `fillna` methods on your dataframe — see https://github.com/pandas-dev/pandas/issues/57734.

Even if you do the recommended fix, you still get this FutureWarning spam when you call your script:
```
FutureWarning: Downcasting object dtype arrays on .fillna, .ffill, .bfill is deprecated and will change in a future version. Call result.infer_objects(copy=False) instead. To opt-in to the future behavior, set `pd.set_option('future.no_silent_downcasting', True)`
```

## Current Behavior
Currently plaguing `grep_more_ioc` calls due to calls to `fillna`

## Possible Solution
Preface these calls with `with pd.option_context('future.no_silent_downcasting', True):` to fix the problem. 

## Steps to Reproduce (for bugs)
Call `fillna` in the latest `pcds_conda` env

aberges-SLAC:
## Description
Small fix to close #224 spamming terminal with the future warnings. Kind of a weird problem for these calls in pandas, take a look at https://github.com/pandas-dev/pandas/issues/57734 if you're interested.

## Motivation and Context
Wanted to fix the warning spam and stay ahead of changes

## How Has This Been Tested?
Have ran `grep_more_ioc` locally on my branch and confirmed the spam goes away and the script works as intended:
image
image

This PR

### DIFF
--- a/scripts/grep_more_ioc.py
+++ b/scripts/grep_more_ioc.py
@@ -492,15 +492,19 @@ def main():
     # pad the disable column based on the grep_ioc output
     if 'disable' not in df.columns:
         df['disable'] = df.index.size*[False]
-    if 'disable' in df.columns:
-        df['disable'] = df['disable'].fillna(False).astype(bool)
+    # handle stupid pandas 3.0 future warnings early
+    with pd.option_context('future.no_silent_downcasting', True):
+        if 'disable' in df.columns:
+            df['disable'] = (df['disable'].infer_objects().fillna(False))
 
     # Fill the NaN with empty strings for rarely used keys
-    for _col in df.columns:
-        if _col not in ['delay']:
-            df[_col] = df[_col].fillna('')
-        else:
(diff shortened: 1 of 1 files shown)


tangkong (approved):
Huh that's annoying.  This seems right to me, though I wish we could just use the proper syntax and not have to revise this if / when we move to pandas 3.0

aberges-SLAC:
> Huh that's annoying. This seems right to me, though I wish we could just use the proper syntax and not have to revise this if / when we move to pandas 3.0

Yeah it's _extremely_ annoying that the only way to make it go away is to opt into the future behavior, even if you avoided down casting in the first place.

PR #227, opened 2024-11-27
drp node bypass for UED machines

laurentroque:
## Description
`restartdaq` command did not function on UED machines as it would request a drp node which was an LCLS-II requirement with a network connection to the drp-network. UED has a stand-a-lone server (ued-cmp002) and no DRP-network. This has been fixed by checking the hostname for the string "ued" before executing a drp check. Similar code has been used in CXI.

## How Has This Been Tested?
DAQ has been restarted through ued-daq computer using the included version of `restartdaq` without issue on 26 November 2024. Tested separately by Ian Roque (iroque) and Patrick Oppermann (opperman).

Comments have been included in `restartdaq`.

### DIFF
--- a/scripts/restartdaq
+++ b/scripts/restartdaq
@@ -124,7 +124,13 @@ else
     PROCMGR="/reg/g/pcds/dist/pds/$HUTCH/current/tools/procmgr/procmgr"
 fi
 
-IS_DAQ_HOST=$(netconfig search "$AIMHOST"-$DAQNETWORK --brief | grep -c $DAQNETWORK)
+# If UED machine; bypass drp node requirement, else; check for drp node
+if [ "$HUTCH" == "ued" ]; then
+    IS_DAQ_HOST=1
+else
+    IS_DAQ_HOST=$(netconfig search "$AIMHOST"-$DAQNETWORK --brief | grep -c $DAQNETWORK)
+fi
+
 if [ "$IS_DAQ_HOST" == 0 ]; then
     HOSTS=$(netconfig search "$HUTCH"-*-$DAQNETWORK --brief | sed s/-$DAQNETWORK//g)
     WORKINGHOSTS=''


patoppermann (approved):
I tested it in UED and RIX both are working with the changes.

scripts/restartdaq
```diff
@@ -124,7 +124,13 @@ else
     PROCMGR="/reg/g/pcds/dist/pds/$HUTCH/current/tools/procmgr/procmgr"
 fi
 
-IS_DAQ_HOST=$(netconfig search "$AIMHOST"-$DAQNETWORK --brief | grep -c $DAQNETWORK)
+# If UED machine; bypass drp node requirement, else; check for drp node
+if [ "$HUTCH" == "ued" ]; then
+    IS_DAQ_HOST=1
```
ZLLentz, line 129: Small question: are there hutch computers other than ued-daq where this could cause an issue?

There's a later (a few lines later) check in the script for IS_DAQ_HOST with the intention of giving suggestions on other hosts to try if you're not on a host that can run the daq. I think it'd be best if any such other hosts would recommend ued-daq to you.

laurentroque: I believe ued-daq is currently the only computer where this may cause an issue.

There are other machines like ued-monitor, but you would not be able to run the daq from that machine.
Do you think it would still be beneficial to have this sort of warning you suggested in case someone on a non-daq machine tries to run `restartdaq`?

ZLLentz: It's probably not the most important thing, but I wonder what you expect/prefer to happen if someone runs this script on ued-monitor or one of the ioc hosts. Just food for thought, I'm not rejecting this PR or anything.

laurentroque: I will add these features in a new pull request

PR #228, opened 2024-12-06
Enh/ued takepeds

silkenelson:
Adding the a script to take pedestals for the UED 360 Hz epix-quad to takepeds
## Description
Check if we are in an LCLS1 hutch (then call the previous scripts) or a new hutch (then for UED, call the new script)

## Motivation and Context
It is easier for UED staff to use commands they use at other instruments,

## How Has This Been Tested?
I checked that this does not break the old hutches in CXI

here

In addition, I have added options to wherepsana to also show information for ami processes.

### DIFF
--- a/scripts/takepeds
+++ b/scripts/takepeds
@@ -15,20 +15,43 @@ if [[($1 == "--help") || ($1 == "-h")]]; then
         exit 0
 fi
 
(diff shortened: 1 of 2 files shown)


vespos:
Looks good to me. I assume that we will change this once we have a more generic takepeds script for LCLS-II?

PR #229, opened 2025-01-30
Fix for UED takepeds

silkenelson:
## Description
Change the environment that needs to be source (necessary due to split environment for LCLS2)

## Motivation and Context
takepeds should work in UED

## How Has This Been Tested?
check that the executable can now be found.

here


patoppermann (approved):
Looks good.

PR #230, opened 2025-01-31
fixes to make makepeds(_psana) work for the UED epix quad (and maybe …

silkenelson:
fixes to make makepeds(_psana) work for the UED epix quad (and maybe also the epix100)

## Description
different extension for xtc files for LCLS2 and LCLS1
arguments for epix-calibration commands have change.

## Motivation and Context
makepeds should work also for UED when using the quad.

## How Has This Been Tested?
running this code on ued1010667, run 33

here.

### DIFF
--- a/scripts/makepeds_psana
+++ b/scripts/makepeds_psana
@@ -435,9 +435,11 @@ if echo "$LCLS2_HUTCHES" | grep -iw "$HUTCH" > /dev/null; then
     
     SIT_ENV_DIR=$SIT_ENV_DIR'/conda2/'
     LCLS2=1
+    XTCEXT='xtc2'
 else
     #echo "This is a LCLS-I experiment"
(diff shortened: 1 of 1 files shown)


scripts/makepeds_psana
patoppermann, line 1: For the archon I used the ffb for xtc-dir?!

silkenelson: Data will not be available there for very long. I think the lcls1 code has a check in there that makes the xtcdir env variable- id have to check if that also takes care of lcls2

patoppermann (approved):
all good

vespos:
The changes are fine for me.
We should probably take the time to split makepeds_psana in makepeds_psana_lcls1 and makepeds_psana_lcls2, and make sure we make the lcls2 version as clean as possible.

silkenelson:
I agree that a split sooner rather than later would be good. My next step was to fix the pedestal plotting code and then go back.

PR #231, opened 2025-02-03
Adding txi to grep_ioc, quoting vars

KaushikMalapati:
## Description

I added txi to the hutches available in grep_ioc
I quoted two variables that caused shellcheck globbing warnings (both for hutchnames which are always one word anyways)

## Motivation and Context

I was trying to use grep_ioc for txi and it said "incorrect hutch name given.  Enter hutch name in lower case (example: xpp)."

## How Has This Been Tested?

I did grep_ioc "id" txi and it gave me all the id lines in the txi/iocmanager.cfg file instead of the above error.

I updated the list of hutches in the README.md entry for grep_ioc and the grep_ioc comment at the top of the file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -269,7 +269,7 @@ optional arguments:<br/>
     <td>
 usage: grep_ioc KEYWORD [hutch]<br/>
         hutch can be any of:<br/>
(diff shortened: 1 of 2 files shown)


ZLLentz (approved):
Looks good to me, good fix!

PR #235, opened 2025-02-09
Quoting variable and replacing echo with printf

KaushikMalapati:
## Description

Instead of echoing nothing, I used printf with newlines
I quoted the wget result where needed because it doesn't make sense to glob archiver queries

I took another look at https://github.com/pcdshub/engineering_tools/pull/204. I think it would be better to do these one by one instead of all at once. 

## Motivation and Context

## How Has This Been Tested?

Interactively

image
image

### DIFF
--- a/scripts/archive-resume
+++ b/scripts/archive-resume
@@ -12,13 +12,12 @@ if [[ -z $1 ]]; then
     usage
     exit 1
 fi
-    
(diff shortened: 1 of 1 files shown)


slactjohnson (approved):
LGTM. Thanks!

PR #236, opened 2025-02-12
add k-line camviewer config to K-line hutches

silkenelson:
add k-line camviewer config to K-line hutches
Also removed wait time functionality as the camviewer redesign changes how timeouts are handled.

## Description
make the txi camviewer work & keep the whole list of cameras for TMO & RIXS

## Motivation and Context
due to TMo& RIXS including comon cameras explicitly, the TXI camviewer was not able to launch IM2K0 due to a found ambiguity
This request fixes this in a quick way.
Really, we should not look for cameras in _all_ config-files, but only the ones included for a given instruments avoiding potential duplicates if e.g. multiple hutch want to call a given camera e.g. SPEC.

## How Has This Been Tested?
I checked the list of available cameras from various instrument machines,


patoppermann (approved):
looks good and should work with txi.
For the future, TXI has to have both beamlines in the camviewer-config.

PR #238, opened 2025-02-14
makepeds for LCLS2

silkenelson:
split makepeds for LCLS1 and LCLS2
add archon support
test/fix epix100 LCLS2 handling
improve error reporting by the script

## Motivation and Context
makepeds should also work for all LCLS2 detectors that need if
It should also return helpful error messages when possible.

## How Has This Been Tested?
re-processe pedestals for selected LCLS1 and LCLS2 experiment/run/detector combos

### DIFF
--- a/scripts/makepeds_psana2
+++ b/scripts/makepeds_psana2
@@ -1,0 +1,594 @@
+#! /bin/bash
+
+# Datasets for testing
+# mfxdaq23,    run 005 : epix100
+# rixx1017523, run 388 : archon
+# uedc00104,   run 021 : epixquad
+
+usage()
+{
+cat << EOF
+usage: $0 options
+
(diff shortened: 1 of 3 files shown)


scripts/makepeds_psana2
```diff
@@ -0,0 +1,594 @@
...
+    case $key in
+        -h|--help)
+            usage
+            exit
+            ;;
+        -A|--alvium)
+            WANT_ALVIUM=1
+            shift
+            ;;
+        -Z|--zyla)
```
patoppermann, line 150: Is this already integrated as a precaution so that we have these 
detectors for HE-XPP?

patoppermann (approved):
Tested in RIX with Archon.
Works and looks good.

PR #241, opened 2025-02-19
Shellcheck eloggrabber

KaushikMalapati:
## Description

In three places, using $() instead of backticks for commands
The rest of changes are just quoting these variables
- HUTCH
- USERNAME
- EXPNAME
- SCRIPT
- $1 (experiment name as argument)
All of which I think are fine to quote and not glob

## Motivation and Context

## How Has This Been Tested?

Interactively, I saw no difference in using the grubber


ZLLentz (approved):
Looks straightforwardly correct to me

PR #243, opened 2025-02-19
add kfe with ssh option

patoppermann:
## Description
Add the option to start kfe as -H (hutch), if you are not on kfe-console it will start a ssh connection.

Please test, for me it's working in my account and rixopr.

```
camViewer -H kfe

HUTCH: kfe
Not in SSH, connecting to kfe-console...
===============================================================================
                                NOTICE TO USERS

This is a Federal computer system and is the property of the United States
Government. It is for authorized use only. Users (authorized or unauthorized)
have no explicit or implicit expectation of privacy.

Any or all uses of this system and all files on this system may be intercepted,
monitored, recorded, copied, audited, inspected, and disclosed to authorized
site, Department of Energy, and law enforcement personnel, as well as
authorized officials of other agencies, both domestic and foreign. By using
this system, the user consents to such interception, monitoring, recording,
copying, auditing, inspection, and disclosure at the discretion of authorized
site or Department of Energy personnel.

Unauthorized or improper use of this system may result in administrative
disciplinary action and civil and criminal penalties. By continuing to use this
system you indicate your awareness of and consent to these terms and conditions
of use. LOG OFF IMMEDIATELY if you do not agree to the conditions stated in
this warning.
===============================================================================
HUTCH:  kfe --
Skipping verify-hutch in remote SSH session
Launching camViewer for kfe.
working with camera of name: IM1K0
/reg/g/pcds/pyps/config/kfe/camviewer/run_viewer.csh --camerapv IM1K0:XTES:CAM --instrument kfe --pvlist /reg/g/pcds/pyps/config/kfe/camviewer.cfg
kfe-console:~> Launching camviewer w/ --camerapv IM1K0:XTES:CAM --instrument kfe --pvlist /reg/g/pcds/pyps/config/kfe/camviewer.cfg from /reg/g/pcds/pyps/config/kfe/camviewer
Log file: /tmp/camviewer.pyw.25-02-19_14:31:02
```

### DIFF
--- a/scripts/camViewer
+++ b/scripts/camViewer
@@ -334,6 +334,8 @@ if [ "$CAMNUM" -eq 0 ]; then
         CAMNAME=cxi_dg1_yag
     elif [ "$hutch" = "mec" ]; then
         CAMNAME=GigE_Questar1
+    elif [ "$hutch" = "lfe" ]; then
+        CAMNAME=xrt_spec
     elif [ "$hutch" = "ued" ]; then
         CAMNAME=ued-gige-01
     else
--- a/scripts/camViewer
+++ b/scripts/camViewer
@@ -181,8 +181,8 @@ EOF
 }
 
 if [[ ($1 == "--help") || ($1 == "-h") ]]; then
-	usage
-	exit 0
+        usage
+        exit 0
 fi
 
 
@@ -202,30 +202,39 @@ NUDGE=0
 
 OPTIONS=$(getopt -o lmredasnc:w:u:H: --long list,main,reboot,enable,disable,acquire,stop,nudge,cam:,rate:,hutch: -n \'$0\' -- "$@")
 
-if [ $? != 0 ] ; then echo "Terminating..." >&2 ; exit 1 ; fi
-
-
+if [ $? != 0 ]; then echo "Terminating..." >&2 ; exit 1 ; fi
 
 eval set -- "$OPTIONS"
-
-
 
 while [ : ]
 do
     case "$1" in
-        -l|--list)  
+        -l|--list)
             LIST=1
             shift
             ;;
         -c|--cam)
             CAM=$2
-            shift 2 ;;
+            shift 2
+            ;;
         -u|--rate)
             RATE=$2
-            shift 2 ;;
+            shift 2
+            ;;
         -H|--hutch)
             hutch=$2
-            shift 2 ;;
+            echo "HUTCH: $hutch" >&2
+            #  If *not* already running in an SSH session, SSH to kfe-console
+            if [[ -z "$IS_REMOTE_SSH" ]] && [ "$hutch" == "kfe" ]; then
+                echo "Not in SSH, connecting to kfe-console..." >&2
+                # Pass all arguments *before* IS_REMOTE_SSH=1
+                ssh -t kfe-console "IS_REMOTE_SSH=1 $DIR/$(basename "$0") \"$@\" ; bash"
+                exit_status=$?
+                echo "SSH exit status: $exit_status" >&2
+                exit 0
+            fi
+            shift 2
+            ;;
         -m|--main)
             MAINSCREEN=1
             shift
@@ -258,9 +267,9 @@ do
             usage
             break
             ;;
-        --) 
-            shift 
-            break 
+        --)
+            shift
+            break
             ;;
         *)
             echo "Internal Error"
@@ -269,6 +278,20 @@ do
     esac
 
 done
+
+# Skip verify-hutch if this is the remote SSH session
+if [ -n "$IS_REMOTE_SSH" ]; then
+   echo "Skipping verify-hutch in remote SSH session" >&2
+else
+    if ! verify-hutch "$hutch"; then
+        hutch=$(get_info --gethutch)
+        if [ "$hutch" == unknown_hutch ]; then
+            echo 'Unknown hutch, specify using -H option with a valid hutch name.'
+            usage
+            exit 1
+        fi
+    fi
+fi
 
 if [ $# -ge 1 ]; then
     echo "Positional arguments unsupported: $@" >&2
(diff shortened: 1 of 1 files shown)


scripts/camViewer
```diff
@@ -202,30 +202,39 @@ NUDGE=0
...
             RATE=$2
-            shift 2 ;;
+            shift 2
+            ;;
         -H|--hutch)
             hutch=$2
-            shift 2 ;;
+            echo "HUTCH: $hutch" >&2
+            #  If *not* already running in an SSH session, SSH to kfe-console
+            if [[ -z "$IS_REMOTE_SSH" ]] && [ "$hutch" == "kfe" ]; then
```
ZLLentz, line 228: It it worth also doing this for lfe and xrt? (to lfe-console)

scripts/camViewer
```diff
@@ -202,30 +202,39 @@ NUDGE=0
...
+                echo "Not in SSH, connecting to kfe-console..." >&2
+                # Pass all arguments *before* IS_REMOTE_SSH=1
+                ssh -t kfe-console "IS_REMOTE_SSH=1 $DIR/$(basename "$0") \"$@\" ; bash"
+                exit_status=$?
+                echo "SSH exit status: $exit_status" >&2
+                exit 0
```
ZLLentz, line 234: Is there a reason we're setting this to zero? Naively I'd assume we should exit with ssh's exit status.

scripts/camViewer
```diff
@@ -202,30 +202,39 @@ NUDGE=0
...
+            ;;
         -H|--hutch)
             hutch=$2
-            shift 2 ;;
+            echo "HUTCH: $hutch" >&2
+            #  If *not* already running in an SSH session, SSH to kfe-console
+            if [[ -z "$IS_REMOTE_SSH" ]] && [ "$hutch" == "kfe" ]; then
+                echo "Not in SSH, connecting to kfe-console..." >&2
+                # Pass all arguments *before* IS_REMOTE_SSH=1
+                ssh -t kfe-console "IS_REMOTE_SSH=1 $DIR/$(basename "$0") \"$@\" ; bash"
```
ZLLentz, line 231: Is there any risk of command-line parameters getting dropped from `$@` thanks to the `shift` calls in the rest of the cli handling here? I suspect we need to cache the original args and pass them unaffected by `shift`.

ZLLentz (commented):
A few minor thoughts
I think this is a good way to circumvent gateway large array shenanigans

scripts/camViewer
```diff
@@ -202,30 +202,39 @@ NUDGE=0
...
 
 eval set -- "$OPTIONS"
 
-
-
 while [ : ]
 do
     case "$1" in
-        -l|--list)  
+        -l|--list)
```
ZLLentz, line 212: One more thought, there are specific argument combinations that should not SSH

For example, if the user does `camViewer -H kfe --list` it should check the list right away, not after ssh-ing to `kfe-console`

ZLLentz: I think simplest would be to relocate the ssh part of this script to the code block below where we've decided to open the camviewer python gui

ZLLentz: For example, one simple way to handle many of my suggestions is to ssh only at the bottom where we invoke the camviewer executable: https://github.com/pcdshub/engineering_tools/blob/ffc6675a8e19a14dc0b7c01e3d884c51f740c9b4/scripts/camViewer#L409-L414

And then we no longer have weird input processing issues or recursive script calls

patoppermann: I relocated the ssh part.

scripts/camViewer
patoppermann: I include lfe, do not see the reason for xrt, it would be the same ssh session.

scripts/camViewer
patoppermann: exit_status=$?
        echo "SSH exit status: $exit_status" >&2
        exit $exit_status

yes, added it.

scripts/camViewer
```diff
@@ -399,17 +414,35 @@ fi
...
+        caput -n "$CAMPVFULL":Acquire 1
     fi
 else
     echo 'cannot check Acquiring PV, try to open camera viewer anyways....'
 fi
 
+if [[ -z "$IS_REMOTE_SSH" ]]; then
+    if [ "$hutch" == "kfe" ]; then
+        echo "Not in SSH, connecting to kfe-console..." >&2
+        ssh -t kfe-console "IS_REMOTE_SSH=1 $EXE --camerapv '$CAMPVFULL' --instrument '$hutch' --pvlist '$PVLIST' --rate '$RATE' ; bash"
```
ZLLentz, line 427: One small detail, `rate` defaults to `1` inside camviewer if it is left as `-1` in this script. I think that's why there's a rate switch below in the non-ssh variant. I'm not sure how much this matters.

References:

- https://github.com/pcdshub/engineering_tools/blob/ffc6675a8e19a14dc0b7c01e3d884c51f740c9b4/scripts/camViewer#L194
- https://github.com/pcdshub/camviewer/blob/dafba34800b613c778124811f0712a2010190f7e/camviewer_ui_impl.py#L541

scripts/camViewer
```diff
@@ -270,6 +269,20 @@ do
 
 done
 
+# Skip verify-hutch if this is the remote SSH session
+if [ -n "$IS_REMOTE_SSH" ]; then
```
ZLLentz, line 273: I believe this variable is no longer used, there are a couple references to it remaining

scripts/camViewer
ZLLentz: Possibly it would make sense to default rate to 5 in this script instead of -1 and then we no longer need to keep track of multiple variants

carogee (approved):
camViewer now loads on the lfe network from my account and as xcsopr.

PR #244, opened 2025-02-20
More error checking ^ reporting

silkenelson:
## Description
We now check the return code of the geometry & other constants deployment scripts

## Motivation and Context
We should let the user now more explicitly when something goes wrong as expecting them to read through many lines of output....

## How Has This Been Tested?
I've tested this does not break anything, but the error scenario has been fixed in the meantime....

### DIFF
--- a/scripts/makepeds_psana
+++ b/scripts/makepeds_psana
@@ -140,7 +140,7 @@ xtcav_dark()
        echo Running "$NJOBS" jobs are PIDS: "${JOBIDS[@]}"
        until check_running_jobs; do
            echo $CALIBTIME' sec have elapsed, checking again in 10 sec'
-           let "CALIBTIME+=10"
+	   CALIBTIME=$((CALIBTIME+10))
            sleep 10
        done
        echo 'All jobs finished'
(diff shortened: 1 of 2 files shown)


scripts/makepeds_psana
```diff
@@ -140,7 +140,7 @@ xtcav_dark()
        echo Running "$NJOBS" jobs are PIDS: "${JOBIDS[@]}"
        until check_running_jobs; do
            echo $CALIBTIME' sec have elapsed, checking again in 10 sec'
-           let "CALIBTIME+=10"
+	   CALIBTIME=$((CALIBTIME+10))
```
patoppermann, line 143: that's the same for me.

patoppermann (approved):
I see more error messages and wait-time, plus clean up. Looks good for me.

ZLLentz:
It's probably ok to ignore the pre-commit shellcheck fwiw

ZLLentz (approved):
I'm not in a position to run this but all the changes look clean

scripts/makepeds_psana
silkenelson: shellchk.....

silkenelson:
Thank you - I've done a few changes suggested by shellchk, but was not exhaustive. Getting closer every time....

PR #250, opened 2025-03-12
add ued .cnf file

patoppermann:
## Description
Ued needs the option to switch between two cnf files.
That's why there is now -u with the options andor.cnf and epix.cnf.


silkenelson (approved):
LGTM

PR #253, opened 2025-03-19
quoting vars, ignoring cd warnings, removing whitespace

KaushikMalapati:
## Description

- disabling three shellcheck warnings about entering /reg/d/iocData/, a specific ioc's directory within that, and the iocInfo directory within that
- quoting the variable i, which is a directory within /reg/d/iocData/ and doesn't need globbing
- quoting KEYWORD in two places, which is the argument passed to grep 

Arguably I shouldn't have ignored the cd into /reg/d/iocData because there's a chance that you are not on a host with the iocData mount. I can add a check for the directory if you think it's a good idea.

## Motivation and Context

## How Has This Been Tested?

Interactively
```
(pcds-5.9.1) [kaushikm@psbuild-rhel7-01 scripts]$ ./grep_pv TST:INTERLOCK
/reg/d/iocData/ioc-tst-alias/iocInfo
TST:INTERLOCK:X:ARTIFICIAL:2, "scalcout"
TST:INTERLOCK:X:ARTIFICIAL:1, "ao"
TST:INTERLOCK:X:ALIAS, "ao"

(pcds-5.9.1) [kaushikm@psbuild-rhel7-01 scripts]$ grep_pv TST:INTERLOCK
/reg/d/iocData/ioc-tst-alias/iocInfo
TST:INTERLOCK:X:ARTIFICIAL:2, "scalcout"
TST:INTERLOCK:X:ARTIFICIAL:1, "ao"
TST:INTERLOCK:X:ALIAS, "ao"

(pcds-5.9.1) [kaushikm@psbuild-rhel7-01 scripts]$ ./grep_pv TST:INT*
/reg/d/iocData/ioc-tst-alias/iocInfo
TST:INTERLOCK:X:ARTIFICIAL:2, "scalcout"
TST:INTERLOCK:X:ARTIFICIAL:1, "ao"
TST:INTERLOCK:X:ALIAS, "ao"

/reg/d/iocData/ioc-tst-try-deploy/iocInfo
TST:INT, "longout"

(pcds-5.9.1) [kaushikm@psbuild-rhel7-01 scripts]$ grep_pv TST:INT*
/reg/d/iocData/ioc-tst-alias/iocInfo
TST:INTERLOCK:X:ARTIFICIAL:2, "scalcout"
TST:INTERLOCK:X:ARTIFICIAL:1, "ao"
TST:INTERLOCK:X:ALIAS, "ao"

/reg/d/iocData/ioc-tst-try-deploy/iocInfo
TST:INT, "longout"

(pcds-5.9.1) [kaushikm@psbuild-rhel7-01 scripts]$ time ./grep_pv TST:INT*
/reg/d/iocData/ioc-tst-alias/iocInfo
TST:INTERLOCK:X:ARTIFICIAL:2, "scalcout"
TST:INTERLOCK:X:ARTIFICIAL:1, "ao"
TST:INTERLOCK:X:ALIAS, "ao"

/reg/d/iocData/ioc-tst-try-deploy/iocInfo
TST:INT, "longout"


real    0m10.107s
user    0m1.220s
sys     0m5.501s
(pcds-5.9.1) [kaushikm@psbuild-rhel7-01 scripts]$ time grep_pv TST:INT*
/reg/d/iocData/ioc-tst-alias/iocInfo
TST:INTERLOCK:X:ARTIFICIAL:2, "scalcout"
TST:INTERLOCK:X:ARTIFICIAL:1, "ao"
TST:INTERLOCK:X:ALIAS, "ao"

/reg/d/iocData/ioc-tst-try-deploy/iocInfo
TST:INT, "longout"


real    0m10.733s
user    0m1.330s
sys     0m5.730s
```

If you tried providing two args to grep_pv, it would use the second as another search location instead of part of the search string which is assume is an unlikely bug since you pvs don't have spaces anyways.
```
[kaushikm@psbuild-rhel7-02 scripts]$ grep_pv "TMO VAC"
grep: VAC: No such file or directory
/reg/d/iocData/ioc-ArbiterPLC/iocInfo
grep: VAC: No such file or directory
IOC.pvlist:PMPS:KFE:PVA_CONN_TMO, "stringin"

grep: VAC: No such file or directory
/reg/d/iocData/ioc-at1k4-gige/iocInfo
grep: VAC: No such file or directory
IOC.pvlist:AT1K4:PPM:CAM:IOC:PVA_CONN_TMO, "stringin"
... (truncated for brevity)
``` 

```
[kaushikm@psbuild-rhel7-02 scripts]$ ./grep_pv "TMO VAC"
[kaushikm@psbuild-rhel7-02 scripts]$
```

### DIFF
--- a/scripts/grep_pv
+++ b/scripts/grep_pv
@@ -16,21 +16,24 @@ fi
 KEYWORD=$1
 
 if [[ -n $KEYWORD ]]; then
+    # shellcheck disable=SC2164
     cd /reg/d/iocData/
-    for i in ioc-*; do 
-	cd /reg/d/iocData/$i
+    for i in ioc-*; do
+	# shellcheck disable=SC2164
+	cd /reg/d/iocData/"$i"
 	if [ -e ./iocInfo/ ] ; then
+	    # shellcheck disable=SC2164
      	    cd iocInfo
 	    if [ -e ./IOC.pvlist ] ; then
-		if grep -q $KEYWORD IOC.pvlist
-		then 
+		if grep -q "$KEYWORD" IOC.pvlist
+		then
 		    pwd
-		    grep $KEYWORD IOC.pvlist;
+		    grep "$KEYWORD" IOC.pvlist;
 		    echo ""
 		else
 		    :
 		fi
-	    fi 
+	    fi
 	fi
 	cd ../../
     done


ZLLentz:
> If you tried providing two args to grep_pv, it would use the second as another search location instead of part of the search string which is assume is an unlikely bug since you pvs don't have spaces anyways.

I agree this was likely a bug in the previous (that shellcheck is now saving us from by mandating the quotes)

scripts/grep_pv
```diff
@@ -16,21 +16,24 @@ fi
 KEYWORD=$1
 
 if [[ -n $KEYWORD ]]; then
+    # shellcheck disable=SC2164
     cd /reg/d/iocData/
```
ZLLentz, line 20: Thinking more critically about the code path here, maybe we can remove this `cd` command entirely instead of disabling shellcheck. This is always overridden by whatever is going on in the loop.

scripts/grep_pv
```diff
@@ -16,21 +16,24 @@ fi
...
-    for i in ioc-*; do 
-	cd /reg/d/iocData/$i
+    for i in ioc-*; do
+	# shellcheck disable=SC2164
+	cd /reg/d/iocData/"$i"
```
ZLLentz, line 23: Perhaps _all_ the cd's in this file aren't actually necessary... we could easily just grep the full path. Do the cds in the loop make this easier to read?

Maybe I'm ambivalent about this in general. It's probably OK for these cds to fail because the next line checks for files.

scripts/grep_pv
```diff
@@ -16,21 +16,24 @@ fi
...
 	if [ -e ./iocInfo/ ] ; then
+	    # shellcheck disable=SC2164
      	    cd iocInfo
 	    if [ -e ./IOC.pvlist ] ; then
```
ZLLentz, line 27: And even with these exist checks, it's not obvious to me what they gain for us compared to letting the grep fail.

ZLLentz (approved):
Anyway I think the changes here are reasonable, if the cds fail the rest still behaves as expected, and we prefer not to exit.

All my confusion is from thinking just a little bit harder about the original implementation.

ZLLentz:
That's fair, I also don't think it matters. We can merge as-is since it's always safe to continue for these specific cds.

PR #261, opened 2025-03-28
add MFX new wave8v3 box at DG1

silkenelson:
## Description
add a new 'box' to the list.

## Motivation and Context
MFX would like to configure the new wave8v3 it has for testing purposes.

## How Has This Been Tested?
This was used as hot-fix before.

here.


vespos (approved):
I assume most changes are the shellcheck script.
Looks good

silkenelson:
Actually, that is me undoing those changes as I only wanted to submit the script as deployed in 'latest'.
Any change you can fix that? This is merging that goes beyond my level (I'd need to create a new branch, check out the newest version & add the other box). Exactly the use case why one should never edited latest (in particular without also doing the right thing)

PR #266, opened 2025-04-04
Improve SLURM integration and command routing for daq utilities

monarin:
## Description

This PR introduces a feature-rich update focused on improving reliability, maintainability, and clarity of the DAQ utility scripts. The core enhancements are:

- ✅ **Slurm command availability check**: Added checks for essential Slurm commands (`squeue`, `sbatch`, `scancel`) to ensure the host environment is valid before executing DAQ-related commands.

- **Unified command dispatch with `checkdaqroute`**: Introduced `checkdaqroute` as a central routing utility for bash scripts, streamlining command routing and reducing redundancy.
  
- 🔁 **Dynamic command dispatch**: Refactored DAQ command handling to use dynamic function lookup, reducing duplication and making it easier to extend or modify supported commands.

- 🧼 **Structured logging + better error handling**: Replaced `print` statements in `daq_utils.py` and `run_daq_utils.py` with `logging`, and improved exception handling to provide clearer failure context.

- 🧩 **Optional `--cnf` and default for `--aimhost`**: Simplified CLI usage by making `--cnf` optional and providing a sensible default for `--aimhost`.

- 🔁 **Updated `isdaqmgr()` to use exit codes**: Changed behavior from string matching to well-defined exit codes for improved shell script integration and reliability.

## Motivation and Context

These changes aim to:

- Improve the reliability of `isdaqmgr` by ensuring it functions correctly across different environments without SLURM dependencies.
- Standardize command dispatch mechanisms to prevent inconsistencies and potential errors in bash scripts.
- Enhance the maintainability and readability of the codebase by reducing dependencies and simplifying command routing.

## How Has This Been Tested?

- Ran DAQ utility scripts in Slurm and non-Slurm environments to verify fallback behavior.
- Confirmed correct exit codes and logging output for all supported subcommands.
- Validated `isdaqmgr()` integration with `checkdaqroute` in bash wrappers like `startdaq`, `stopdaq`, and `restartdaq`.

- Code-level docstrings and comments updated
- Structured commit messages provided for each atomic change

### DIFF
--- a/scripts/daq_utils.py
+++ b/scripts/daq_utils.py
@@ -8,7 +8,6 @@ import logging
 import os
 import socket
 import subprocess
-import sys
 import time
 from subprocess import PIPE
 
@@ -122,15 +121,6 @@ class DaqManager:
             self.hutch,
             self.cnf_file,
         )
-
-    def isdaqmgr(self):
-        result = self.hutch in DAQMGR_HUTCHES
-        msg = "true" if result else "false"
-        logging.debug("isdaqmgr: %s", msg)
-        if result:
-            sys.exit(0)
-        else:
-            sys.exit(1)
 
     def isvaliduser(self):
         valid = getpass.getuser() == self.user
--- a/scripts/daq_utils.py
+++ b/scripts/daq_utils.py
@@ -180,32 +180,42 @@ class DaqManager:
         return daq_host
 
     def calldaq(self, subcmd, daq_host=None):
-        prog = "daqmgr"
-        cmd = f"pushd {self.scripts_dir} > /dev/null; "
-        cmd += f'source {os.path.join(self.scripts_dir, "setup_env.sh")}; '
-        cmd += f"WHEREPROG=$(which {prog}); set -x; "
-        cmd += f"$WHEREPROG {subcmd} {os.path.join(self.scripts_dir, self.cnf_file)}; {{ set +x; }} 2>/dev/null; "
-        cmd += "popd > /dev/null;"
-
+        # Determine where to run the command
         if subcmd == "restart":
             query_daq_host = self.wheredaq()
         else:
             query_daq_host = self.wheredaq_internal()
 
         if daq_host is None:
-            daq_host = query_daq_host
-            if daq_host is None:
-                daq_host = LOCALHOST
+            daq_host = query_daq_host or LOCALHOST
 
-        logging.debug("Executing command on host: %s", daq_host)
+        logging.debug("Executing DAQ command on host: %s", daq_host)
+
+        # Build daq commad
+        env_setup = os.path.join(self.scripts_dir, "setup_env.sh")
+        try:
+            get_daqmgr_cmd = f"source {env_setup}; which daqmgr"
+            daqmgr_path = subprocess.check_output(get_daqmgr_cmd, shell=True, stderr=PIPE).strip().decode("utf-8")
+        except subprocess.CalledProcessError as e:
+            logging.error("Failed to locate 'daqmgr' after sourcing env: %s", e.stderr.decode())
+            return
+
+        daq_command = f"{daqmgr_path} {subcmd} {os.path.join(self.scripts_dir, self.cnf_file)}"
+        print(f"Running command: {daq_command}")
+
+        cmd = f"""
+pushd {self.scripts_dir} > /dev/null
+source {env_setup}
+{daq_command}
+popd > /dev/null
+"""
         if daq_host == LOCALHOST:
             try:
-                ret = subprocess.run(
-                    cmd, stdout=PIPE, stderr=PIPE, shell=True, check=True
-                )
-                logging.debug("Command output: %s", ret.stdout.decode())
+                ret = subprocess.run(cmd, stdout=PIPE, stderr=PIPE, shell=True, check=True)
+                logging.debug("Command stdout:\n%s", ret.stdout.decode())
+                logging.debug("Command stderr:\n%s", ret.stderr.decode())
             except subprocess.CalledProcessError as e:
-                logging.error("Local command failed: %s", e.stderr.decode())
+                logging.error("Local command failed:\n%s", e.stderr.decode())
         else:
             try:
                 call_sbatch(cmd, daq_host, self.scripts_dir)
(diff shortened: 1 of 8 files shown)


ZLLentz:
I really appreciate your detailed PR description here, it makes it really easy for someone like me to understand what and why even though I don't interface with slurm at all

ZLLentz:
The pre-commit checks are advisory here, not compulsory, and do not have to slow down a merge here.
The only rule is "one passing review", preferably from someone who has used the previous version or knows what's going on.
The standard of review is something like "Is this documented well? Does this make good forward progress without introducing issues?"
If people who know what's going on aren't available I'm happy to review too.

ZLLentz:
If we did want to address the pre-commit, which we don't have to:
- The first one should be ignored, we can formally ignore it with a little comment (I'm happy to come back later and do this)
- The second one is clearly a flake8 bug, not a problem with the code here. It's parsing the shell command as if it was python... I can come back through and fix this too (unrelated to the PR)

monarin:
> The pre-commit checks are advisory here, not compulsory, and do not have to slow down a merge here. The only rule is "one passing review", preferably from someone who has used the previous version or knows what's going on. The standard of review is something like "Is this documented well? Does this make good forward progress without introducing issues?" If people who know what's going on aren't available I'm happy to review too.

Thanks! That context is really helpful.

This PR was mainly motivated by two issues that Silke ran into:

**No clear check for SLURM availability**: There wasn’t a reliable way to detect whether the node running `daqmgr` actually supports Slurm commands like `squeue`, `sbatch`, etc. This led to confusing failures when these commands were missing.

**Error handling and messaging**: Error handling in `run_daq_utils.py` and `daq_utils.py` was inconsistent—some failures produced stack traces or silent failures instead of helpful messages. This PR introduces more structured logging and clearer user-facing errors.

Suggested Testing Scenarios
We can test the core functionality with the following:

**Non-daqmgr hutch or `--aimhost` usage**: Running DAQ commands in this mode should fall back to the existing non-daqmgr (legacy) behavior and bypass SLURM-related logic.

**Daqmgr hutch behavior**:

Normal case: Commands should execute as expected.

Simulated failure: Temporarily pause slurmd or remove access to SLURM binaries. The commands should now fail cleanly with informative error messages (instead of crashing or hanging).

We could also explore additional edge cases covered by this PR if time allows, but I think validating these two paths covers the essential functionality.

**Future Consideration**
Going forward, for changes like this, I think it would be great to have a production-like environment that we can switch to for testing. That way, hutch scientists could also try out the updated commands themselves in addition to our internal testing. What do you think?

ZLLentz:
> Going forward, for changes like this, I think it would be great to have a production-like environment that we can switch to for testing. That way, hutch scientists could also try out the updated commands themselves in addition to our internal testing. What do you think?

I think this is a really good idea. I'm not sure how I could help facilitate this (since I don't know how things are deployed now on the server-side), but in the context of this specific repo we can absolutely deploy updates in a limited/dev capacity and give it to specific to try.

Thank you for the detailed notes, I'll give you my best go at a review

scripts/daq_utils.py
```diff
@@ -15,185 +15,212 @@
...
-            daq_host = self.wheredaq(quiet=True)
+            daq_host = self.wheredaq_internal()
             if daq_host is not None:
-                if self.isdaqmgr(quiet=True):
+                if self.hutch in DAQMGR_HUTCHES:
                     for i_retry in range(MAX_RETRIES):
-                        daq_host = self.wheredaq(quiet=True)
+                        daq_host = self.wheredaq_internal()
                         if daq_host is None:
+                            logging.debug("DAQ has stopped after %d retries.", i_retry)
```
ZLLentz, line 141: Note for the future: It's probably totally good for now, but if we end up adding even more if/indent chains here it gets kind of hard to read (I blame Python!), so in the future we should consider breaking some of these bits out into helper functions to keep it readable.

ZLLentz (approved):
I'm :+1: on merging from a docs/code quality perspective and you've tested it yourself (I'm not sure how I would run/test it as me)

Do you want to merge as-is or do you want to turn it over for user testing first? I don't mind either way since it's not so bad come back and revise again in a new PR.

monarin:
> I'm 👍 on merging from a docs/code quality perspective and you've tested it yourself (I'm not sure how I would run/test it as me)
> 
> Do you want to merge as-is or do you want to turn it over for user testing first? I don't mind either way since it's not so bad come back and revise again in a new PR.

Thanks so much for the quick review and support—really appreciate you taking the time to go through the changes! 🙏

I think it makes sense to wait a bit and give Silke a chance to weigh in, especially since she initially requested the changes. That way, she can get a clear view of where things stand and let us know if anything else is needed before we move forward.

I checked the schedule, and the next good opportunity for a release and potential user testing looks like during PAMM, April 14–16. That might be a great window to roll this out and gather feedback from hutch scientists.

Happy to hold off for now and then merge either after Silke signs off or as we get closer to that window. Thanks again for helping push this forward!

ZLLentz:
Sounds good! I'll click the button to request a review from Silke.

ZLLentz:
I think we missed the previously mentioned window- maybe we should just merge it?

monarin:
> I think we missed the previously mentioned window- maybe we should just merge it?

Yes, I think it’s okay to go ahead and merge it.

The only thing I didn’t test directly from the hutch side is the error message when SLURM isn’t available on the node. I was able to test that scenario on other nodes where I could safely simulate the failure, and the error handling worked as expected.

On TMO and RIX, there are base processes running, so it wasn’t practical to stop `slurmd` there to fully reproduce the condition.

Also—sorry I didn’t get to this during the PAMM window as originally planned. I got tied up with other work and couldn’t follow through in time. That said, aside from that one edge case, everything else should behave as before, just with improved logging and error handling.

Thanks again for reviewing and helping move this forward!

ZLLentz:
>  I got tied up with other work and couldn’t follow through in time

Me too! There's so much going on!
Merging

scripts/restartdaq
patoppermann, line 1: `-C (LCLS2 DAQMGR Hutches ONLY!) Select a cnf to use.`

not used in this PR anymore.

PR #268, opened 2025-04-17
MNT: switch off of deprecated os build and straight to latest

ZLLentz:
This version of the ubuntu runner image is being removed. We've chosen to go straight to `latest` everywhere to hopefully not need to manually switch it again.

PR #269, opened 2025-04-18
ENH: ioc deploy rebuild

ZLLentz:
## Description

Add a `rebuild` command to `ioc-deploy`.
This will:
- strip any existing write permissions
- make
- apply write permissions

Which will be useful when trying to build a bunch of common IOCs on both rhel7 and rocky9.

I'm considering also adding a hosts argument to ioc-deploy to deploy to multiple architectures all at once, but I've deferred this because:
- The complexity of this "just one simple deploy script" is getting pretty high
- I didn't want to deal with auth error handling at EOD

Possibly the next update for `ioc-deploy` will bring it into an externally tracked module and then I can add complexity without causing issues in engineering_tools.

## Motivation and Context

https://jira.slac.stanford.edu/browse/ECS-7695

>ioc-deploy works great but has some annoying behavior when you want to build on multiple architectures.
>
> You have to:
>
> - pick rhel7 or rocky9 to start and then run ioc-deploy
> - run the command to un-write-protect the release
> - pick the other os, cd to the deployment and make
> - run the command to re-write-protect the release
>
> I'd prefer to streamline this a bit to either:
>
> - a re-deploy sort of command that handles steps 2-4 all on its own
> - some smarts or options in the initial command to build on both rhel7 and rocky9

## How Has This Been Tested?

Interactively by deploying and redeploying `ioc-common-pdu_snmp` to `~zlentz/temp/ioc` (this is the only tagged IOC with rocky9 compatibility right now).

I'll update the readme.

### DIFF
--- a/scripts/ioc_deploy.py
+++ b/scripts/ioc_deploy.py
@@ -1120,7 +1120,7 @@ def rearrange_sys_argv_for_subcommands() -> str:
     be interpreted the same as:
     ioc-deploy update-perms ro -p some_path
 
-    Otherwise, the first example here is interpretted as if -p was never passed,
+    Otherwise, the first example here is interpreted as if -p was never passed,
     which could be confusing.
 
     Returns
--- a/scripts/ioc_deploy.py
+++ b/scripts/ioc_deploy.py
@@ -20,14 +20,14 @@ to rebuild an existing release on a new architecture or remove it entirely.
 
 Example command:
 
-"ioc-deploy -n ioc-foo-bar -r R1.0.0"
+"ioc-deploy -n ioc-common-foo -r R1.0.0"
 
 This will clone the repository to the default ioc directory and run make using the
 currently set EPICS environment variables, then apply write protection.
 
 With default settings, this will clone
-from https://github.com/pcdshub/ioc-foo-bar
-to /cds/group/pcds/epics/ioc/foo/bar/R1.0.0
+from https://github.com/pcdshub/ioc-common-foo
+to /cds/group/pcds/epics/ioc/common/foo/R1.0.0
 then cd and make and chmod as appropriate.
 
 If the repository exists but the tag does not, the script will ask if you'd like
@@ -41,10 +41,10 @@ if this is more convenient for you.
 
 Example commands:
 
-"ioc-deploy update-perms rw -n ioc-foo-bar -r R1.0.0"
-"ioc-deploy update-perms ro -n ioc-foo-bar -r R1.0.0"
-"ioc-deploy update-perms rw -p /cds/group/pcds/epics/ioc/foo/bar/R1.0.0"
-"ioc-deploy update-perms ro -p /cds/group/pcds/epics/ioc/foo/bar/R1.0.0"
+"ioc-deploy update-perms rw -n ioc-common-foo -r R1.0.0"
+"ioc-deploy update-perms ro -n ioc-common-foo -r R1.0.0"
+"ioc-deploy update-perms rw -p /cds/group/pcds/epics/ioc/common/foo/R1.0.0"
+"ioc-deploy update-perms ro -p /cds/group/pcds/epics/ioc/common/foo/R1.0.0"
 
 The rebuild action will run make again on your current OS.
 It will conveniently temporarily remove write protections from the release for
@@ -53,8 +53,8 @@ Like update-perms, you can invoke this using similar commands as the deploy acti
 
 Example commands:
 
-"ioc-deploy rebuild -n ioc-foo-bar -r R1.0.0"
-"ioc-deploy rebuild -p /cds/group/pcds/epics/ioc/foo/bar/R1.0.0"
+"ioc-deploy rebuild -n ioc-common-foo -r R1.0.0"
+"ioc-deploy rebuild -p /cds/group/pcds/epics/ioc/common/foo/R1.0.0"
 """
 
 import argparse
--- a/scripts/ioc_deploy.py
+++ b/scripts/ioc_deploy.py
@@ -278,6 +278,7 @@ def args_to_dataclass(args: argparse.Namespace) -> CliArgs:
     Therefore, it's the duty of this function to fill in any
     missing arguments with their default values.
     """
+    # TODO use dataclasses.replace once we're free of py3.6
     kw = vars(DEFAULT_ARGS)
     kw.update(vars(args))
     return CliArgs(**kw)
--- a/scripts/ioc_deploy.py
+++ b/scripts/ioc_deploy.py
@@ -54,7 +54,7 @@ Like update-perms, you can invoke this using similar commands as the deploy acti
 Example commands:
 
 "ioc-deploy rebuild -n ioc-foo-bar -r R1.0.0"
-"ioc-deploy rebuild -p -p /cds/group/pcds/epics/ioc/foo/bar/R1.0.0"
+"ioc-deploy rebuild -p /cds/group/pcds/epics/ioc/foo/bar/R1.0.0"
 """
 
 import argparse
@@ -73,6 +73,8 @@ EPICS_SITE_TOP_DEFAULT = "/cds/group/pcds/epics"
 GITHUB_ORG_DEFAULT = "pcdshub"
 CHMOD_SYMLINKS = os.chmod in os.supports_follow_symlinks
 PERMS_CMD = "update-perms"
+REBUILD_CMD = "rebuild"
+ALL_SUBCOMMANDS = (PERMS_CMD, REBUILD_CMD)
 
 logger = logging.getLogger("ioc-deploy")
 
@@ -114,7 +116,7 @@ else:
     DeployInfo = SimpleNamespace
 
 
-def get_parser(subparser: bool = False) -> argparse.ArgumentParser:
+def get_parser(subparser: str = "") -> argparse.ArgumentParser:
     current_default_target = str(
         Path(os.environ.get("EPICS_SITE_TOP", EPICS_SITE_TOP_DEFAULT)) / "ioc"
     )
@@ -147,8 +149,20 @@ def get_parser(subparser: bool = False) -> argparse.ArgumentParser:
         type=force_lower,
         help="Select whether to make the deployment permissions read-only (ro) or read-write (rw).",
     )
+    # rebuild_parser doesn't have unique arguments but let's be consistent in our approach
+    rebuild_parser = subparsers.add_parser(
+        REBUILD_CMD,
+        help=(
+            f"Use 'ioc-deploy {REBUILD_CMD}' to help rebuild write-protected releases. "
+            f"See 'ioc-deploy {REBUILD_CMD} --help' for more information."
+        ),
+        description=(
+            "Rebuild a deployment, even if it is write protected. "
+            "This will briefly relax write permissions, run make, and then reapply permission restrictions."
+        ),
+    )
     # shared arguments
-    for parser in main_parser, perms_parser:
+    for parser in main_parser, perms_parser, rebuild_parser:
         parser.add_argument(
             "--name",
             "-n",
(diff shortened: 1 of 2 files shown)


janeliu-slac (dismissed):
(Just noticed the failing check) Logically, the code looks good to me. So, approving whenever the failing check passes.

ZLLentz:
Thank you Jane!

I'd like at least two brains to review because this script has consequences for ecs devops sorts of things

tangkong:
This is on my radar, I'll take a look soon

ZLLentz:
I think I just need to merge #268 into this branch and the CI will run again here

scripts/ioc_deploy.py
```diff
@@ -1071,29 +1144,50 @@ def rearrange_sys_argv_for_subcommands():
 
     Otherwise, the first example here is interpretted as if -p was never passed,
     which could be confusing.
+
```
tangkong, line 1147: interpretted -> interpreted above, but I can't comment on it since there was no diff up there.

I think it's probably good to leave a really clear note that this mutates sys.argv in place.  As a function that isn't passed any arguments, I'd probably not expect any side effects past the return value.  

The name of the function is clear, but I just want to make sure we're extra clear here

scripts/ioc_deploy.py
```diff
@@ -349,6 +389,33 @@ def main_perms(args: CliArgs) -> int:
...
+    rval = _perms(deploy_dir=deploy_dir, allow_write=True, dry_run=args.dry_run)
+    if rval != ReturnCode.SUCCESS:
+        return rval
+    rval = make_in(deploy_dir=deploy_dir, dry_run=args.dry_run)
+    if rval != ReturnCode.SUCCESS:
+        logger.error(f"Nonzero return value {rval} from make")
+        return rval
+    rval = _perms(deploy_dir=deploy_dir, allow_write=False, dry_run=args.dry_run)
+    if rval == ReturnCode.SUCCESS:
+        logger.info("Rebuild complete!")
```
tangkong, line 415: Nit: So if we error out on this `_perms` call, we don't tell the user exactly what went wrong?  Even though we get loud in `main`, the message is a generic "deploy errored out"

scripts/ioc_deploy.py
```diff
@@ -323,9 +353,19 @@ def main_perms(args: CliArgs) -> int:
...
+
+    return _perms(deploy_dir=deploy_dir, allow_write=allow_write, dry_run=args.dry_run)
+
+
+def _perms(deploy_dir: str, allow_write: bool, dry_run: bool) -> int:
+    """
+    Helper function: re-usable user-facing core of main_perms.
+
+    For example, this is used in both main_perms and in main_rebuild.
+    """
```
tangkong, line 365: Should this try-except block just be incorporated into `set_permissions`?  That function has basically the same signature as this one.

tangkong (approved):
I'm only now getting a chance to really scrutinize the argparse setup here.  It does what you want (allows the subcommand to be placed anywhere in the arg-list), but I do wonder if dynamically choosing which parser to use is the simplest way around this. 

I am able to get weird incantations like this to work, but that's probably a feature more than anything: 
```bash
./ioc-deploy --dry-run -n ioc-tst-vacuum update-perms -r R1.0.0 ro
```

I have no constructive suggestion in this vein, so this is more of a comment than a critique.

I took this for a spin and it looks like the rebuild functionality works as advertised.  (through dry runs at least)

Some nits on the internals of this but nothing that should hold up merging if rushed

ZLLentz:
> I do wonder if dynamically choosing which parser to use is the simplest way around this

> I am able to get weird incantations like this to work

Both of these points are valid. The weird incantations are the point but I'm invested in cleaning up how messy this ends up being.

scripts/ioc_deploy.py
ZLLentz: I'll do these

scripts/ioc_deploy.py
ZLLentz: I was hoping that _perms itself would have interesting output and logging, I'll double-check

scripts/ioc_deploy.py
ZLLentz: You're right, I'll refactor a bit.

scripts/ioc_deploy.py
```diff
@@ -1114,14 +1208,12 @@ def _main() -> int:
...
-            do_perms_cmd = False
-        if do_perms_cmd:
-            rval = main_perms(args)
-        else:
+        if not subcmd:
             rval = main_deploy(args)
+        elif subcmd == PERMS_CMD:
+            rval = main_perms(args)
+        elif subcmd == REBUILD_CMD:
+            rval = main_rebuild(args)
```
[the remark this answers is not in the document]
ZLLentz: I'll investigate incorporating this to simplify

scripts/ioc_deploy.py
```diff
@@ -1071,29 +1144,50 @@ def rearrange_sys_argv_for_subcommands():
...
-    except IndexError:
-        return
+    subc_index = None
+    for subcommand in ALL_SUBCOMMANDS:
+        try:
+            subc_index = sys.argv.index(subcommand)
+        except ValueError:
+            ...
+        else:
+            break
```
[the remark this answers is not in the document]
ZLLentz: This is worth doing, and raising if it happens.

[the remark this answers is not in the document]
tangkong: I tried this, I thought that because the parser only expected one positional argument that the cli would throw an error early

Basically I thought it was impossible to pass multiple parsers through to this point

```bash
$ ./ioc-deploy --dry-run -n ioc-tst-vacuum -r R1.0.0 update_perms ro rebuild
usage: ioc-deploy [-h] [--version] [--name NAME] [--release RELEASE] [--ioc-dir IOC_DIR] [--path-override PATH_OVERRIDE] [--auto-confirm] [--dry-run] [--verbose] [--github_org GITHUB_ORG]
                  {update-perms,rebuild} ...
ioc-deploy: error: unrecognized arguments: update_perms ro
```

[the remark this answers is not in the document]
ZLLentz: Ah, I see. Maybe the other changes discussed here could at least give us a better error though.

scripts/ioc_deploy.py
ZLLentz: I've confirmed that we get a clear error messages whenever we set the return code to non-success values.

ZLLentz: I'm going to refactor a bit more to not return return codes from subfunctions, raising is more useful and is handled in main appropriately

scripts/ioc_deploy.py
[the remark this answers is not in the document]
ZLLentz: I'm trying out this and other ideas to figure out if the cli parsing in general can be cleaner here

scripts/ioc_deploy.py
```diff
@@ -43,6 +45,16 @@
...
 "ioc-deploy update-perms ro -p /cds/group/pcds/epics/ioc/foo/bar/R1.0.0"
+
+The rebuild action will run make again on your current OS.
+It will conveniently temporarily remove write protections from the release for
+the duration of the make so you don't have to do this in multiple steps.
+Like update-perms, you can invoke this using similar commands as the deploy action.
+
+Example commands:
+
+"ioc-deploy rebuild -n ioc-foo-bar -r R1.0.0"
```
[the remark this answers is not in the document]
ZLLentz: Yes, good call out. Using something that unambiguously represents an IOC name in the docs here would make them more clear.

Possibly the can't-find-it-on-github error needs to be more clear too?

ZLLentz:
I had a few more ideas I tried for cleaning up the argument parsing, I didn't like them. I was going to ask for a merge without those (and just to do the docs update mentioned just above this comment) but I have one more idea to try that I came up with today.

If that option fails I'll probably just switch to `click` when this moves from `engineering_tools` to a dedicated repo where I can properly specify dependencies.

scripts/ioc_deploy.py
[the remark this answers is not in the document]
ZLLentz: I am ultimately opting against the set_defaults(func=) pattern because I don't like it (after trying it out).

I mainly dislike how the resulting code is less clear, as `main` will be running some arbitrary function assigned at the parser, instead of a specific choice between multiple functions based on the parser output.

scripts/ioc_deploy.py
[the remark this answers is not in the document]
ZLLentz: I've decided to use `ioc-common-foo`. I almost used `ioc-common-gigECam` but then I decided I wanted to use a definitely fake IOC name so people who paste the examples don't cause havok in the gigECam repositories.

ZLLentz:
I think this new version of the parser is about as simple as it can get.

The problem with the old parser is that default values from the optional arguments in the subparsers e.g. `rebuild` would overwrite user-provided values to the base parser.

E.g. the args are processed like
- create blank namespace
- fill blank namespace with base parser arguments and defaults
- overwrite namespace with subparser arguments and defaults

So the subparser defaults would always win, hence me moving the subparsers earlier in sys.argv as a hack

This new version prevents default values from being written to the namespace at all, so default arguments from the subparser no longer overwrite user arguments to the main parser.

The new arg processing is:
- create blank namespace
- fill blank namespace with base parser arguments, but only defaults from base-parser-only arguments, arguments shared with the subparsers are fully omitted
- repeat with the subparsers, only fill in default args for arguments unique to that subparser
- after argument parsing, include any totally missing defaults as appropriate

scripts/ioc_deploy.py
```diff
@@ -180,35 +220,61 @@ def get_parser(subparser: bool = False) -> argparse.ArgumentParser:
...
+
+    This is required because our subparsers share arguments
+    with the main parser, and if they have default values
+    they will override the main parser every time.
+
+    Therefore, it's the duty of this function to fill in any
+    missing arguments with their default values.
+    """
+    kw = vars(DEFAULT_ARGS)
+    kw.update(vars(args))
```
tangkong, line 276: Nit: I think [`dataclass.replace()`](https://docs.python.org/3/library/dataclasses.html#dataclasses.replace) could work here.  Not that this is that much more verbose.

scripts/ioc_deploy.py
```diff
@@ -135,13 +162,25 @@ def get_parser(subparser: bool = False) -> argparse.ArgumentParser:
...
-    for parser in main_parser, perms_parser:
+    for parser in main_parser, perms_parser, rebuild_parser:
         parser.add_argument(
             "--name",
             "-n",
             action="store",
-            default="",
+            default=argparse.SUPPRESS,
             help=(
                 "The name of the repository to deploy. This is a required argument. "
```
tangkong, line 185: https://docs.python.org/3/library/argparse.html#required

Though python suggests we don't, we can specify this argument as required if we want it to be.

tangkong: Same with `--release`

tangkong (approved):
This looks good to me on the whole, just a small nit.

Just noting that we've given up the freedom to place the sub-command arg wherever we want, but that makes more sense to me personally

ZLLentz:
> Just noting that we've given up the freedom to place the sub-command arg wherever we want, but that makes more sense to me personally

You're right. The new constraint is that the base parser only args (`--version`, `--github-org`). Must be given to the base parser. Maybe I should just include these in the subparser loop.

scripts/ioc_deploy.py
ZLLentz: til dataclass.replace

ZLLentz:
> Maybe I should just include these in the subparser loop.

After looking again, I'm OK with github org only being provided to the base parser, because that's the only action that connects to github.

Same with version, I prefer ioc-deploy rebuild --version to fail (not that --version means a lot for engineering_tools here)

scripts/ioc_deploy.py
ZLLentz: forgot that I'm still supporting rhel7 installed python3 which doesn't have dataclasses yet
putting this in as a comment instead of as a change

ZLLentz:
I think this is the version I want to merge and move on from

tangkong (approved):
All for moving on 👍

PR #270, opened 2025-04-18
MNT: replace flake8 + isort with ruff

ZLLentz:
## Description

I'm putting this up for consideration because I find this tool very helpful.

Replace the flake8 pre-commit with ruff, using the config I've been using.
The settings are negotiable (comparing with the existing .flake8 config file, maybe there are more changes to make here before merging)

## Motivation and Context

flake8 doesn't fix things for you
isort is slow

ruff is fast and fixes things for you

## How Has This Been Tested?

This is the config I've been using in other projects

n/a

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -2,7 +2,7 @@
 # See https://pre-commit.com/hooks.html for more hooks
 repos:
 -   repo: https://github.com/pre-commit/pre-commit-hooks
-    rev: v4.4.0
+    rev: v5.0.0
     hooks:
     -   id: no-commit-to-branch
     -   id: trailing-whitespace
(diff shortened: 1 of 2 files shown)


tangkong:
How does this work when people submit code that does not conform to style guidelines?  (as in, the submitter did not run pre-commit locally)  Ruff might format stuff in the CI runner, but we don't have a way to translate that back into commits right?

As long as ruff throws an error in the pre-commit job when it fixes things the workflow is the same, but if it fixes things up silently we'll have to adjust our jobs to let the submitter know there are changes to be made.

ZLLentz:
All pre-commit tasks, ruff format included, have a nonzero return code when they edit files. So, it will cause the job to fail and the diff will appear in the output.

PR #271, opened 2025-04-25
MFX is also a LCLS2 hutch for makepeds

silkenelson:
add MFX to list of LCLS2 hutches for makepeds.

### DIFF
--- a/scripts/makepeds
+++ b/scripts/makepeds
@@ -200,7 +200,7 @@ else
     HUTCH=${EXP:0:3}
 fi
 
-LCLS2_HUTCHES="rix, tmo, txi, ued"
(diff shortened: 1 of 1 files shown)


silkenelson:
Not a good reason that txi exists in makepeds, but not takepeds - in practice neither of them are needed for TXI right now. I have another addition to the script that it meant to catch a failure I managed to run into today that is rare, so I'll go make this consistent. Actually, takepeds could use the same mechanism that restartdaq etc use as this is run on the PCDS systems and not S3DF.

silkenelson:
The other comment was also very good - that must have been how I got MFX to work. Obviously I could not remember this, so the new way is cleaner and I've removed the --lcls2 flag in an upcoming release.

PR #276, opened 2025-05-01
Shellcheck daq waitwin

KaushikMalapati:
## Description

- Using -gt instead of > for an arithmetic comparison
- Using $() instead of legacy backticks in several places
- Removing trailing whitespace
- Quoting variables
- Making SEARCHOPT an array

## Motivation and Context


ZLLentz (approved):
Seems pretty straightforward, good cleanup

ZLLentz:
Do you think we should try it on a control room machine then? Or maybe the makerspace desktops?

PR #285, opened 2025-05-09
Restore -C config flag and add -v; suppress ShellCheck warnings in restartdaq

monarin:
## Description
This PR restores the previously removed `-C`/`--cnf` configuration flag and adds support for a `-v`/`--verbose` flag for consistency across DAQ utilities. It also defers the config file existence check until the full path is available. Additionally, this update includes a workaround to suppress ShellCheck SC2034 warnings for unused variables `DAQMGR_CNF` and `VERBOSE`, which are required for compatibility with downstream tools.

## Motivation and Context
These flags are necessary for proper argument parsing across multiple DAQ scripts. The missing `-C` option was causing failures when passing configuration files. The ShellCheck warnings were false positives—these variables are used indirectly via downstream utilities—so they are now referenced explicitly to keep pre-commit checks clean.

## How Has This Been Tested?
Tested manually by running `restartdaq` with various combinations of options including `-C` and `-v`. Verified that configuration is passed correctly and logging behaves as expected. Also confirmed that the updated code passes `pre-commit` including `shellcheck`.

Comments were added inline near the ShellCheck suppression directive. Argument help strings were also updated or confirmed to be clear for users running `--help`.

### DIFF
--- a/scripts/restartdaq
+++ b/scripts/restartdaq
@@ -32,7 +32,7 @@ if [[ ($1 == "--help") || ($1 == "-h") ]]; then
 fi
 
 CORESIZE=0
-while getopts "m:pwscdu:" OPTION
+while getopts "m:pwscdCu:v" OPTION
 do
     case $OPTION in
 	p)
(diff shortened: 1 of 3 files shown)


scripts/restartdaq
```diff
@@ -64,6 +70,10 @@ do
 	esac
 done
 
+# Prevent ShellCheck SC2034 warning for unused vars passed to other scripts
+# shellcheck disable=SC2034
+: "$DAQMGR_CNF" "$VERBOSE"
```
ZLLentz, line 75: Can someone explain what this means to me?
To my eye, shellcheck's warning here is correct. This line is a no-op and these variables are both unused and not exported. So, how are they being passed to other scripts?
Is there a bash trick I don't know?

PR #292, opened 2025-05-13
ENH/MNT: updates/cleanups to conda setup scripts for py312

### LINKED ISSUE #202: Include "dev" in prompt when sourcing dev_conda
## Expected Behavior

```
$ source dev_conda
(dev-pcds-5.9.1)
```

## Current Behavior

```
$ source dev_conda
(pcds-5.9.1)
```

## Possible Solution

`CONDA_PROMPT_MODIFIER` environment variable

## Context

I spied this env variable today and remembered this old idea/suggestion/request and decided to write it down

ZLLentz:
## Description

- Prepare for `pcds-6.0.0`, which will be python 3.12
- Clean up the duplicated code between `pcds_conda` and `dev_conda` and put it in a shared script.
- Add `dev-` to the conda prompt when using `dev_conda` (closes #202)
- Add variables that tell us which script we sourced for downstream consumers (like me in my crazy prompt) to use.
 
## Motivation and Context

We're gearing up for an open beta release of `pcds-6.0.0` which will be python 3.12.

This update enables you to access that environment via:
```
export PCDS_CONDA_VER=6.0.0
source pcds_conda
```
Python 3.9 is still the default.

While I was here I remembered #202 and found a simple way to include it for normal users.

For users with custom prompts etc. like me I added two new exported variables to let us know you're using pcds conda or dev conda in case we want to use this information.

## How Has This Been Tested?

Interactively only

It will be announced when the env is deployed
I updated the readme but it's somewhat unrelated

### DIFF
--- a/scripts/condarc.yaml
+++ b/scripts/condarc.yaml
@@ -1,0 +1,4 @@
+# Customizations for dev-conda
+# Some options can only be set using config files
+# Included in dev_conda using the CONDARC environment variable
+env_prompt: '(dev-{default_env}) '
--- a/scripts/dev_conda
+++ b/scripts/dev_conda
@@ -36,5 +36,6 @@ PYDM_DISPLAYS_PATH="$(cut -d ":" -f 3- <<< "${PYDM_DISPLAYS_PATH}")"
 export PYDM_DISPLAYS_PATH
 
 # In case you want to use this info
-export IS_PCDS_CONDA=0
+export IS_PCDS_CONDA=1
+export IS_TAG_CONDA=0
 export IS_DEV_CONDA=1
--- a/scripts/dev_conda
+++ b/scripts/dev_conda
@@ -4,7 +4,8 @@ usage()
 cat << EOF
 usage: source $0
 
-Source this to activate the latest version a pcds conda environment.
+Source this to activate a shared in-development version of a pcds conda environment.
+The environment provided by this is not guaranteed to be stable.
 EOF
 }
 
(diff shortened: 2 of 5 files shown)


scripts/condarc.yaml
```diff
@@ -0,0 +1,4 @@
+# Customizations for dev-conda
+# Some options can only be set using config files
+# Included in dev_conda using the CONDARC environment variable
+env_prompt: '(dev-{default_env}) '
```
tangkong, line 4: 🚀

scripts/dev_conda
```diff
@@ -13,46 +14,27 @@ if [[ ($1 == "--help") || ($1 == "-h") ]]; then
...
-TOKEN_DIR="${CONDA_ROOT}/.tokens"
-source "${TOKEN_DIR}/typhos.sh"
-
-# Revert to Software Raster when SSH to avoid issues with QtQuick.
-# The same was done with PyDM and this fixes Designer and friends.
-if [ -n "${SSH_CONNECTION}" ]; then
-  export QT_QUICK_BACKEND="software"
-fi
+# In case you want to use this info
+export IS_PCDS_CONDA=0
```
tangkong, line 39: So interestingly, I always thought of dev-conda as a patched pcds_conda, but still pcds_conda.
I think it's good to differentiate the tagged from dev conda, but the base is still pcds_conda.  

I think this is fine as is, maybe `IS_TAG_PCDS_CONDA` would have put my mind more at ease but this isn't really an issue

scripts/shared_conda_setup.sh
```diff
@@ -0,0 +1,131 @@
...
+# Clear externally set paths that may mess with imports etc.
+unset LD_LIBRARY_PATH
+unset PYTHONPATH
+
+# Pick up EPICS environment variable settings just in case user did not
+source "${SETUP_ROOT}/epics-ca-env.sh"
+
+# BASE_ENV transition points
+FIRST_PCDS_PY38="4.0.0"
+FIRST_PCDS_PY39="4.2.0"
```
tangkong, line 24: Wow did we really not bump the major version for a python version change?

tangkong (approved):
This LGTM, simple enough of a refactor.

scripts/dev_conda
ZLLentz: Ah I think I agree fully, I'm not sure why I picked 0

ZLLentz: Maybe something like this is best then:
```
export IS_PCDS_CONDA=1
export IS_TAG_CONDA=1
export IS_DEV_CONDA=0
```
```
export IS_PCDS_CONDA=1
export IS_TAG_CONDA=0
export IS_DEV_CONDA=1
```

scripts/shared_conda_setup.sh
ZLLentz: Yeeep

tangkong (approved):
LGTM

scripts/dev_conda
ZLLentz: I think we can make our prompt mods or any other downstream consumers work with any of these variants, but maybe what I'm going for here is intuitive verbiage

Maybe I've spent too many brain minutes on this already

tangkong: Sorry to bring it up 😆

PR #293, opened 2025-05-15
revertdaq

silkenelson:
The commit reverts changes to the DAQ related scripts implemented after R3.13.4
## Description
There were significant changes to the DAQ related code that broke a fair number of scripts. 

## Motivation and Context
There were significant changes to the DAQ related code that broke a fair number of scripts. The plan is to re-introduce the desired changes slowly w/ detailed test.

## How Has This Been Tested?
calling DAQ related scripts in CXI, RIX & TXI. 
MFX has weird implementations of engineering tool in their hutch python, so this will be tested later.
UED still needs to finalize moving to a SLURM DAQ based on python files.

### DIFF
--- a/scripts/daq_utils.py
+++ b/scripts/daq_utils.py
@@ -3,56 +3,50 @@
 #
 # Do not use non-standard python.
 ####################################################
+import errno
 import getpass
-import logging
 import os
 import socket
 import subprocess
-import sys
 import time
 from subprocess import PIPE
 
-DAQMGR_HUTCHES = ["tmo", "rix", "txi", "mfx"]
+DAQMGR_HUTCHES = ["tmo", "rix", "txi", "ued"]
 LOCALHOST = socket.gethostname()
 SLURM_PARTITION = "drpq"
 SLURM_JOBNAME = "submit_daq"
+SLURM_SCRIPT = "submit_daq.sh"
 SLURM_OUTPUT = "slurm_daq.log"
 MAX_RETRIES = 10
 
 
-def call_subprocess(*args):
+def silentremove(filename):
     try:
-        output = subprocess.check_output(args, stderr=PIPE).strip()
-        return output.decode("utf-8")
-    except subprocess.CalledProcessError as e:
-        logging.error("Subprocess call '%s' failed with error: %s", " ".join(args), e)
-        raise
+        os.remove(filename)
+    except OSError as e:
+        if e.errno != errno.ENOENT:  # errno.ENOENT = no such file or directory
+            raise
+
+
+def call_subprocess(*args):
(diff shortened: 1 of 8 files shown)


ZLLentz:
Reviewing this is kind of hard because it's a reversion: the point is that we're going back, so yes we do lose some nice things but we plan to bring it back later.

Documentation is clear and PR is well-motivated
The test notes are helpful to know which hutches this does and does not apply to

For review I'll at least note places in the original code we're re-introducing if I think something is strange, otherwise as I said previously it's hard to be diligent here.

scripts/daq_utils.py
```diff
@@ -3,232 +3,197 @@
...
 def call_sbatch(cmd, nodelist, scripts_dir):
     sb_script = "#!/bin/bash\n"
-    sb_script += f"#SBATCH --partition={SLURM_PARTITION}\n"
-    sb_script += f"#SBATCH --job-name={SLURM_JOBNAME}\n"
-    sb_script += f"#SBATCH --nodelist={nodelist}\n"
-    sb_script += f"#SBATCH --output={os.path.join(scripts_dir, SLURM_OUTPUT)}\n"
+    sb_script += f"#SBATCH --partition={SLURM_PARTITION}" + "\n"
+    sb_script += f"#SBATCH --job-name={SLURM_JOBNAME}" + "\n"
+    sb_script += f"#SBATCH --nodelist={nodelist}" + "\n"
+    sb_script += f"#SBATCH --output={os.path.join(scripts_dir, SLURM_OUTPUT)}" + "\n"
```
ZLLentz, line 40: It's at least a bit strange to pull the `"\n"` out of the fstrings. The resulting string is the same but there's an extra step now.

scripts/restartdaq
```diff
@@ -6,13 +6,9 @@ cat << EOF
 usage: $0 options
 
 OPTIONS:
--w sort windows after start
--d dss-test setup
--p select partition (same as used last)
--s silent (do not email jana)
--c enable core files
--C (LCLS2 DAQMGR Hutches ONLY!) Select a cnf to use.
--u UED .cnf file
```
ZLLentz, line 15: Was nobody using the options we're dropping here?

ZLLentz (approved):
With the caveats above and my two small notes/questions, I don't see any reason to hold this up.

vespos:
This has been tested in several hutches, right?

silkenelson:
I've added Vincent change, but do want to point out that this is NOT the right solution. We should figure out WHY get_info now breaks in certain cases.

scripts/restartdaq
silkenelson: They were either not used or have been rewritten to also allow for a longer command line option as I moved away from getopts.

PR #295, opened 2025-05-19
Enh/revertdaq - MFX is treated as LCLS2 instrument & cleanup

silkenelson:
## Description
During the initial 'revertdaq' pull request MFX became an LCLS1 hutch again. While they use that DAQ setup currently, we have discussed that they should be considered an LCLS2 hutch as far as engineering_tools goes.
At the same time, I have also removed a second definition of LCLS2 instruments in the checkcnf script.

## Motivation and Context
As discussed above.

## How Has This Been Tested?
I have used MEC, MFX & TXI to make sure the scripts I changed still work

unfortunately, just here.

### DIFF
--- a/scripts/checkcnf
+++ b/scripts/checkcnf
@@ -15,15 +15,14 @@ if [[ ($1 == "--help") || ($1 == "-h") ]]; then
 fi
 
 HUTCH=$(get_info --gethutch)
+DAQTYPE=$(get_info --daq)
 
-LCLS2_HUTCHES="rix, tmo, ued"
-if echo "$LCLS2_HUTCHES" | grep -iw "$HUTCH" > /dev/null; then
-    # shellcheck disable=SC1090
-    source /reg/g/pcds/dist/pds/"$HUTCH"/scripts/setup_env.sh
-    PROCMGR='procmgr'
-else
-    PROCMGR="/reg/g/pcds/dist/pds/$HUTCH/current/tools/procmgr/procmgr"
+if [[ $DAQTYPE == 'LCLS2' ]]; then
+    echo 'LCLS2 no longer uses cnf files!'
+    exit 0
 fi
+
+PROCMGR="/reg/g/pcds/dist/pds/$HUTCH/current/tools/procmgr/procmgr"
 
 CNFEXT=.cnf
 CNFFILE="$HUTCH"$CNFEXT
@@ -38,7 +37,7 @@ fi
 STATUS=$($PROCMGR status /reg/g/pcds/dist/pds/"$HUTCH"/scripts/"$CNFFILE" control_gui)
 ISOK=$(echo "$STATUS" | grep -c platform)
 if [[ $ISOK == 0 ]]; then
-    echo The cnf file for "$HUTCH" can be parsed
+    echo Good, the cnf file for "$HUTCH" can be parsed.
 else
     echo WARNING: the cnf file for "$HUTCH" cannot parsed!
 fi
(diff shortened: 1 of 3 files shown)


scripts/takepeds
```diff
@@ -23,20 +23,17 @@ SIT_ENV_DIR="/cds/group/psdm/sw"
 
 HUTCH=$(get_info --gethutch)
 #UED is still using the LCLS2 DAQ w procmgr
```
patoppermann, line 25: that should now be out of date.

scripts/checkcnf
```diff
@@ -15,16 +15,15 @@ if [[ ($1 == "--help") || ($1 == "-h") ]]; then
 fi
 
 HUTCH=$(get_info --gethutch)
+DAQTYPE=$(get_info --daq)
```
ZLLentz, line 18: I like using this script instead of hard-coding the hutches.
Note that it fails sometimes, maybe it's not important here but:

<details>

```
zlentz@xpp-control:~$ get_info --daq
Traceback (most recent call last):
  File "/reg/g/pcds/engineering_tools/R3.13.8/scripts/get_info.py", line 217, in <module>
    exp = resp.json().get("value", {}).get("name")
  File "/cds/group/pcds/pyps/conda/py36/envs/pcds-3.4.0/lib/python3.6/site-packages/requests/models.py", line 898, in json
    return complexjson.loads(self.text, **kwargs)
  File "/cds/group/pcds/pyps/conda/py36/envs/pcds-3.4.0/lib/python3.6/site-packages/simplejson/__init__.py", line 525, in loads
    return _default_decoder.decode(s)
  File "/cds/group/pcds/pyps/conda/py36/envs/pcds-3.4.0/lib/python3.6/site-packages/simplejson/decoder.py", line 370, in decode
    obj, end = self.raw_decode(s)
  File "/cds/group/pcds/pyps/conda/py36/envs/pcds-3.4.0/lib/python3.6/site-packages/simplejson/decoder.py", line 400, in raw_decode
    return self.scan_once(s, idx=_w(s, idx).end())
simplejson.errors.JSONDecodeError: Expecting value: line 1 column 1 (char 0)
```

</details>

ZLLentz (approved):
I think this is good forward progress and the test matrix of mec/mfx/txi seems sufficient.

I assume ued is similar enough to txi to not need its own test.

I'm curious what the correct way to test `takepeds` is, it's not mentioned in the PR description but I assume it's become pretty routine to you.

scripts/takepeds
silkenelson: Yes, the code is gone, but the comment has remained...

scripts/checkcnf
silkenelson: I think this is because I'm looking for files present for the 'current' experiment and XPP is in 'shutdown' now. Unfortunately, this will also fail before the first run has been taken. I should make a note to catch those problems and at least return why the script does not work.
There is an alternative to use the list in Mona's daqmgr-script so at least only have _one_ list - for uses 'online' either will work, but for anything where data exists, the get_info approach is more correct as this will correctly reflect if MFX was LCLS1 or LCLS2 for a given experiment.
I'm open for feedback if I should switch to the daqmgr-check approach as checkcnf is really an 'online' script.

ZLLentz (approved):
Still curious about how one tests `takepeds`
Still happy to merge, probably one of the CDSOs should look too.

patoppermann:
We have to test takepeds in the hutches with known good detectors, I will test RIX tomorrow. At the moment, we have 24/7 beamtime.

ZLLentz:
Was closing this without merging intentional?

silkenelson:
I suspect I was positive this work _had_ been merged and I wanted to clean up branches as the forking-process makes my life hard if I keep too many things around after I no longer need them. I am no super sure what to do now but thankfully all the code changes are still available here so that I could redo the work if I don't find a local version

PR #297, opened 2025-05-29
MFX is treated as LCLS2 instrument & cleanup

ZLLentz:
Restore deleted #295

@silkenelson


silkenelson (approved):
Yes, that looks like what I had before I messed things up!

PR #298, opened 2025-06-04
Enh mp2 loop

silkenelson:
Quite some cleanup - started w/ shellcheck fixes
## Description
the code should also work when you have multiple of the same detector type

## Motivation and Context
the code did not deal with/ two epix100.

## How Has This Been Tested?
on a UED runs as well as a run in MFX

### DIFF
--- a/scripts/makepeds_psana2
+++ b/scripts/makepeds_psana2
@@ -442,7 +442,6 @@ if [[ $HAVE_EPIX100 -ge 1 ]]; then
 	    echo '$MYDET is a $DETTYPE, looking for epix100'
 	    continue
 	fi
-	    
         echo "now calibrate $MYDET"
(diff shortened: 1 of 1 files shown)


gadorlhiac (approved):
I think the logic changes look correct. You should pick up the detectors either from the type or name (or both) and then discard matches that have the incorrect type. This should still work in the case where the detector name no longer matches the type.

Can be tested in MFX with:
```bash
/sdf/home/s/snelson/git_engtools/engineering_tools/scripts/makepeds_psana2 --queue milano --run 10 --experiment  mfx100852324
```

The other changes are mostly improved quoting and formatting.

spellcheck is still unhappy about a number of things. I'll leave it to others to decide on whether these need to be addressed before merging.

PR #299, opened 2025-06-05
same loop fix as for makepeds_psana2

silkenelson:
The looping over several epix10k or jungfrau detectors has been broken for a while. This pull request fixes this.


gadorlhiac (approved):
Looks correct to me. Similar to the fix a day or two ago for the LCLS2 version.

PR #304, opened 2025-07-17
Server stat expert edt 

silkenelson:
## Description
This pull request contains mostly changes to the 'expert' option of serverStat. 
stopdaq had a few lines removed that were no longer necessary and makepeds_psana no longer requests 5 CPUs where 1 suffices.
Other minor changes to the information displayed have been added and three other minor changes to other script have been packaged

## Motivation and Context
serverStat --expert
could lead to machines becoming unresponsive - which is not the expected behavior for a command that is simply supposed to show you some information about said server. The only recent change has been an update to the EDT code that was called to gain more information about those cards. Given that we have essentially discontinued the use of these cards, I have chosen to simple display that there _is_ one.

## How Has This Been Tested?
processed pedestals for the epixquad in MEC & the jungfrau 4M in MFX
ran serverStat ioc-xrt-rec06 expert without adverse effects
(host w/ EDT card)

here


ZLLentz (approved):
The changes seem well-motivated to me and are strict improvements

PR #305, opened 2025-07-21
Shellcheck stopdaq and wheredaq

KaushikMalapati:
## Description

- Removed trailing whitespace in stopdaq
- Fixed braces in wheredaq awk (https://www.shellcheck.net/wiki/SC1083)
- Quoted CNFFILE in stopdaq (file name)
- In both, quoted $(basename "$0") argument to daqutils (evaluates to stopdaq or wheredaq respectively)
- In both, disabled SC2068 for $@ argument(s) to daqutils (https://www.shellcheck.net/wiki/SC2068). Not sure if there are any globbed arguments you would intentionally provide, like a path to multiple cnf files.

### DIFF
--- a/scripts/stopdaq
+++ b/scripts/stopdaq
@@ -3,7 +3,8 @@
 # Check if the current's user hutch uses daqmgr. If so,
 # use the python utilities to manage the daq.
 if [ "$(daqutils isdaqmgr)" = "true" ]; then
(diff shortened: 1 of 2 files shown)


silkenelson:
I would test wheredaq & stopdaq together anyways. I would test in:

silkenelson (approved):

PR #307, opened 2025-07-22
Trailing whitespace removal and quoting vars in ami_offline_psana

KaushikMalapati:
## Description

- Removed trailing whitespace
- Used -r flag to read in experiment
- Using double brackets for REBIN (1 or empty), ami_path (filepath), and NOPLUG (1 or empty)
- Quoting entire echo expression
- Disabling SC2086 for ami execution line. We cannot quote ARGSTR because it it has multiple arguments. It could be made an array but I thought this was easier

### DIFF
--- a/scripts/ami_offline_psana
+++ b/scripts/ami_offline_psana
@@ -44,7 +44,7 @@ if [ -z "$EXP" ]; then
     echo getting the experiment "$EXP"
     HUTCH=$(get_hutch_name)
     if [ "$HUTCH" == 'unknown_hutch' ]; then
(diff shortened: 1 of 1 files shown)


ZLLentz (commented):
The diff looks good.

For testing this one, I don't quite remember what is done- is it just picking some known experiment and trying it out? Does it have to be run on a specific machine? The real meat of the code is held externally in `offline_ami` so there probably isn't that much to test here if the application launches.

ZLLentz (approved):
My intuition is that this is sufficiently thorough testing :+1:

PR #308, opened 2025-07-23
configdb whitespace and shellcheck

KaushikMalapati:
## Description

- Removed trailing whitespaced
- Replaced legacy backticks with $() notation
- Quoted these variables - opt_dir, USER, EXP, HUTCH, EXE
- Assigning ret_val2
- Added -r flag to read username
- Removed unused ARGSTR variable

## How Has This Been Tested?

Interactively, opens the window and gives the same cli output as the last version of this

### DIFF
--- a/scripts/configdb_readxtc
+++ b/scripts/configdb_readxtc
@@ -8,7 +8,7 @@ We will run configdb_readxtc
 
 OPTIONS:
 -u user (needs to be able to log into the psananeh/feh)
--e expnumber 
+-e expnumber
 EOF
 }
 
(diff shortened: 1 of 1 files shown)


silkenelson:
This only works for LCLS1 and you need to run it on S3DF - I don't remember if I updated this to work work on S3DF (paths may need to be different etc), so there is a chance it actually does not work. So try on S3DF on a recent XCS or CXI experiment - that has the highest chance of working. My vague recollection is that the last times I used it I change to the correct files in the GUI.

silkenelson:
I'll approve this. Looking for data under /reg/d/... is not particularly useful, but that is not part of this cleanup.

PR #309, opened 2025-07-24
Quoting vars, using $() instead of backticks

KaushikMalapati:
## Description

- Spaces between brackets and if condition
- Using $() instead of `` notation
- Removing the unused variable LCLS2
- Quoting the variables SIT_ENV_DIR (a path), HUTCH, station (a number), and RUN (run number)
- Disabling warning for indirect error code check, since the command executed depends on whether it's LCLS 1 or 2 and the hutch. I could have saved the command based on the branches taken and executed in the if statement to check the exit code directly, but chose not to.

### DIFF
--- a/scripts/takepeds
+++ b/scripts/takepeds
@@ -10,12 +10,12 @@ Takes a run with dark images for use in pedestals, and posts to the elog.
 EOF
 }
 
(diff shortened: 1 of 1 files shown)


ZLLentz:
The diff looks clean
This does something like take a daq run and save some configuration information, right? We'll need a real system to test on.

PR #310, opened 2025-07-24
FIX: imgr, iocmanager when hutch is passed explicitly

ZLLentz:
## Description

- Fix an issue where the imgr and iocmanager wrapper scripts didn't launch properly when any specific hutch was chosen.
  - Launching for "the current hutch I'm in" was unaffected.
- Ensure hutch is lowercase

## Motivation and Context

I'd previously missed this in my testing. Ilia was trying to use these as documented and ran into an issue where the parser was passing the hutch name twice to the underlying iocmanager parser.

In the future when everyone updates to the R3 series these scripts will want some revisions to simplify the handling even further.

## How Has This Been Tested?

I re-ran these scripts as multiple hutches with various input arguments and everything behaved as expected.

Here only

### DIFF
--- a/scripts/imgr
+++ b/scripts/imgr
@@ -33,11 +33,11 @@ fi
 
 # For newer IOC manager, use the built-in help and env setup
 if [[ -f "${IOCMANAGER}"/scripts/imgr ]]; then
-  if [[ "$HUTCH" == "unknown_hutch" ]]; then
+  if [[ ("$HUTCH" == "unknown_hutch") || (-n $ARGHUTCH) ]]; then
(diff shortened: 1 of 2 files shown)


ZLLentz:
I promise this worked great in testing

ZLLentz:
I'd like to get this in alongside the other thing we want to tag today if possible

tangkong (approved):
LGTM

tangkong:
I saw this and had a moment of dejavu, but it turns out I had just peeked at the diff last week and not left a review

ZLLentz:
I'm planning to tag/release engineering tools tomorrow when I have more brainpower to announce the update from #311

PR #314, opened 2025-08-06
MNT: restore background launch behavior (&) for iocmanager R3+

ZLLentz:
## Description

Add `&` to the iocmanager launcher script to make the GUI launch in the background instead of the foreground.

## Motivation and Context

The old version of this script had this behavior and people like it.

## How Has This Been Tested?

Interactively with tst and txi

Here only


baljamal (approved):
LGTM

PR #328, opened 2025-12-09
MNT: remove references to netconfig from engineering_tools

ZLLentz:
## Description

Remove all references to netconfig and include suitable replacement calls to sdfconfig or host

- `restartdaq`: Replace the `netconfig search *-fez` sort of call with the equivalent `sdfconfig` search for fez subnets (can't do the same search because the -fez etc. are part of the same entry as the main interface)
- `serverStat`: All calls that used `netconfig` to search for hostname existence have been replaced by the `host` utility, which is faster and wont disappear. (Incidentally: `serverStat status` is now much faster). The `netconfig` call that was used to view the netconfig output has been replaced by the equivalent `sdfconfig` call.

## Motivation and Context

https://jira.slac.stanford.edu/browse/ECS-9486

Netconfig is going away
The replacement tool is sdfconfig
In some cases, we can get away with using neither, so that is preferred when possible for forward compatibility.

## How Has This Been Tested?

Interactively
- `restartdaq` I've forced the bash script into this error state and made sure it gave reasonable results
- `serverStat` was run through the paces with the `status` and `expert` commands on servers with/without fez to get some output variety

Here only

### DIFF
--- a/scripts/restartdaq
+++ b/scripts/restartdaq
@@ -88,21 +88,21 @@ cd /reg/g/pcds/dist/pds/"$HUTCH"/scripts/ || exit
 DAQNETWORK='fez'
 PROCMGR="/reg/g/pcds/dist/pds/$HUTCH/current/tools/procmgr/procmgr"
 
-if [ "$IS_DAQ_HOST" == 0 ]; then
-    # There is currently no suitable way to find daq-compatible hosts in sdfconfig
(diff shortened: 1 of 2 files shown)


slactjohnson (approved):
This looks good to me. I really like the use of DNS over `netconfig`/`sdfconfig` where possible.

PR #331, opened 2026-01-29
ENH: add xrt to camviewer script, don't ssh if you don't need to

ZLLentz:
## Description

- Add `xrt` as a valid option for `camViewer`
- Add `xrt` as a valid hutch for `verify-hutch`
- Make `xrt` ssh you to `lfe-console` to open the gui
- If you're already on the hutch we need to ssh to, don't ssh

## Motivation and Context

@patoppermann asked me about ssh-ing to avoid bad gateway performance for camViewer, this is his code from last year but extended to allow xrt
Small cleanup while here

## How Has This Been Tested?

Interactively only

Here only

### DIFF
--- a/scripts/camViewer
+++ b/scripts/camViewer
@@ -386,15 +386,15 @@ else
     echo 'cannot check Acquiring PV, try to open camera viewer anyways....'
 fi
 
-current_host="$(hostname)"
 if [[ -z "$IS_REMOTE_SSH" ]]; then
-    if [ "$hutch" == "kfe" ] && [ "$current_host" != "kfe-console" ]; then
+    current_host="$(hostname)"
+    if [[ "$hutch" == "kfe" && "$current_host" != "kfe-console" && "$current_host" != "xbdo-control" ]]; then
         echo "Not in SSH, connecting to kfe-console..." >&2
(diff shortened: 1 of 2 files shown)


scripts/camViewer
silkenelson, line 1: Is checking for the console machines still sufficient? Don't we need to include the XBDO machine in 901 in this list? Ideally, we'd check for the network interfaces instead of listing machines, but that could be an improvement pull request for later.

silkenelson (commented):
As in the comment - I'd like to make sure we are dealing ok with the XBDO machine otherwise the machine-name approach is fine for now.

scripts/camViewer
ZLLentz: You're right, this could do a little more without much extra effort

Let's include just the xbdo machine for now since that one definitely doesn't need an ssh, putting in another name is trivial, and it will have immediate helpful impact.

Later we can check our interface list as you say- this would mostly be for running camViewer on ioc machines at this point, but it would future-proof us for certain configuration changes/additional consoles

silkenelson (approved):
Well, this fell off my radar.

ZLLentz:
Mine too

PR #332, opened 2026-01-30
startami: strip trailing '})' from amicmd to avoid eval syntax error

laurentroque:
## Description
Strip trailing config-artifact characters (e.g., `})`) from the `amicmd` string constructed in `scripts/startami` before it is executed.

## Motivation and Context
`startami -c cxi_0.cnf` could fail with:
`eval: syntax error near unexpected token ')'`

On CXI hosts, `amicmd` was being parsed from the `ami_client` line in `cxi_0.cnf` and could end up with a trailing `})`, causing bash `eval` to error. This change sanitizes the constructed command to avoid the syntax error.

## How Has This Been Tested?
- `cxi-monitor`: `/cds/home/i/iroque/programs/engineering_tools/scripts/startami -c cxi_0.cnf` (reaches expected restart prompt; no eval syntax error)
- `cxi-control`: `/cds/home/i/iroque/programs/engineering_tools/scripts/startami -c cxi_0.cnf` (successfully launches `online_ami`; no eval syntax error)
- `xcs-control`: `/cds/home/i/iroque/programs/engineering_tools/scripts/startami` (successfully launches online_ami; no eval syntax error)

- Jira ticket: [ECS-9707](https://jira.slac.stanford.edu/browse/ECS-9707)

### DIFF
--- a/scripts/startami
+++ b/scripts/startami
@@ -71,9 +71,7 @@ ami_path=$ami_base_path$(grep ami_GUI_path /reg/g/pcds/dist/pds/"$HUTCH"/scripts
 
 proxy_cds=$(/reg/g/pcds/dist/pds/"$HUTCH"/current/tools/procmgr/procmgr status /reg/g/pcds/dist/pds/"$HUTCH"/scripts/"$CONFIG" | grep ami_proxy | awk '{print $1}' | sed s/'-fez'/''/g)
 
-amicmd=$(grep ami_client /reg/g/pcds/dist/pds/"$HUTCH"/scripts/"$CONFIG" | grep -v '#' |  awk 'BEGIN { FS = ":" }; { print $4}' | sed s/ami_GUI_path//g | sed s/\'+proxy_cds/"$proxy_cds"/g | sed  s:\'+expname:"$EXPNAME"/:g | sed s/+\'//g | sed s/\'\}\)//g)
-# Strip trailing cnf/python artifacts (e.g. "})") so eval doesn't choke
-amicmd=$(printf "%s" "$amicmd" | sed 's/[[:space:]]*[})][})]*$//')
+amicmd=$(grep ami_client /reg/g/pcds/dist/pds/"$HUTCH"/scripts/"$CONFIG" | grep -v '#' |  awk 'BEGIN { FS = ":" }; { print $4}' | sed s/ami_GUI_path//g | sed s/\'+proxy_cds/"$proxy_cds"/g | sed  s:\'+expname:"$EXPNAME"/:g | sed s/+\'//g | sed s/\'\}\)//g | sed 's/[})][})]*$//')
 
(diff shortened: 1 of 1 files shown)


scripts/startami
```diff
@@ -72,6 +72,8 @@ ami_path=$ami_base_path$(grep ami_GUI_path /reg/g/pcds/dist/pds/"$HUTCH"/scripts
 proxy_cds=$(/reg/g/pcds/dist/pds/"$HUTCH"/current/tools/procmgr/procmgr status /reg/g/pcds/dist/pds/"$HUTCH"/scripts/"$CONFIG" | grep ami_proxy | awk '{print $1}' | sed s/'-fez'/''/g)
 
 amicmd=$(grep ami_client /reg/g/pcds/dist/pds/"$HUTCH"/scripts/"$CONFIG" | grep -v '#' |  awk 'BEGIN { FS = ":" }; { print $4}' | sed s/ami_GUI_path//g | sed s/\'+proxy_cds/"$proxy_cds"/g | sed  s:\'+expname:"$EXPNAME"/:g | sed s/+\'//g | sed s/\'\}\)//g)
+# Strip trailing cnf/python artifacts (e.g. "})") so eval doesn't choke
+amicmd=$(printf "%s" "$amicmd" | sed 's/[[:space:]]*[})][})]*$//')
```
ZLLentz: Some nitpicky questions, if this works I think it's correct to merge without addressing these.
I'm mostly trying to update my own sed/regex knowledge.

- Why is this a new command using `printf` instead of joining onto the `| sed` train on line 74?
- (Conversely, if new commands are preferred for readability, maybe we should unwrap more of them)
- Why `printf` with no formatting over `echo`?
- What's `[[:space:]]*` doing in the sed here? Are you expecting e.g. `   })`? Your PR description only mentions `})`
- It looks like you're using a literal `[})]` followed by a `[})]*`, but in regex we have a + operator for matching 1 or more instances of a symbol. This could have been `[})]\+`

laurentroque: OK, one `sed` call on line 74 is probably more efficient. I will change this.

laurentroque: To Zach's second point, Line 74 is not too daunting once read carefully, but during a mid-experiment quick-fix it was just easier to add another line rather than risk messing up the previous line.

laurentroque: On why use `printf` over `echo`, this was just was found via StackOverflow first, I have no real opinion on it. If I fold this command into the previous `sed` line then we don't need to think about this too hard.

laurentroque: On point 4, it looks like the :space: part is not needed. I will remove and re-test.
On point 5, I was unaware of the + operator in regex.

laurentroque:
@ZLLentz your comments have now been addressed and incorporated in the latest push to my fork. The `printf` command is now folded into the original `sed` command.

Script has been re-tested as done in this Pull Request's "**How Has This Been Tested?**" section.

ZLLentz (approved):

PR #335, opened 2026-05-27
ENH: add setup script for ctrlenv

ZLLentz:
## Description

This is a script for opting into the new pixi/uv environments.
Sourcing the script sets some shared environment variables and adds the `ctrlenv-pathmunge`, ~`ctrlenv-activate`~ (delayed to a later PR), and `ctrlenv-versions` bash functions to your shell.

To use these environments:
```
source ctrlenv_setup.sh
ctrlenv-pathmunge
```
Would be the most basic invocation. This adds the `latest-released` set of applications to your path. There are also options for adding a specific release or a specific environment version to your path, or adding a specific single application- just specify what you want e.g.
```
ctrlenv-pathmunge pytmc/v2.20.0
```

I expect most people to just the most basic setup and never think about it again, but these options need to be provided.

To check what versions are available, you can use `ctrlenv-versions`:

```
ctrlenv-versions pytmc
```

## Motivation and Context

I'd like have a way for people to use the production-level deployments of external tools like pixi and uv and internal tools like pytmc without needing to activate pcds_conda/dev_conda, and I'd like it to be usable this week. Long-term I want this to be the way we distribute python environments.

Since `pcds_conda` and `dev_conda` are both in this repo, this repo seems like the right place to include this too.

See https://jira.slac.stanford.edu/browse/ECS-10013 for more.

## How Has This Been Tested?

Interactively- all the commands I've tried have worked as I intended.

Added a lengthy docstring to the script here.
I will add an entry to the repo README

### DIFF
--- a/scripts/ctrlenv_setup.sh
+++ b/scripts/ctrlenv_setup.sh
@@ -137,10 +137,14 @@ ctrlenv-pathmunge() {
             if [ -d "${CTRLENV_ENVS}/${tg}/bin/${arch}" ]; then
                 pathmunge "${CTRLENV_ENVS}/${tg}/bin/${arch}"
                 return 0
-            else
-            # Wasn't a version- pick latest
+            elif [ -d "${CTRLENV_ENVS}/${tg}/latest-released/bin/${arch}" ]; then
+                # Wasn't a version- pick latest
                 pathmunge "${CTRLENV_ENVS}/${tg}/latest-released/bin/${arch}"
                 return 0
+            else
+                # Error state: let's try to be a little bit helpful
+                echo "Found matching env series ${CTRLENV_ENVS}/${tg} but missing one of arch or version match or no latest-released" >&2
+                return 1
             fi
         fi
     done
@@ -149,10 +153,14 @@ ctrlenv-pathmunge() {
         if [ -d "${CTRLENV_APPS}/${target}/bin/${arch}" ]; then
             pathmunge "${CTRLENV_APPS}/${target}/bin/${arch}"
             return 0
-        else
-        # Wasn't a version- pick latest
+        elif [ -d "${CTRLENV_APPS}/${target}/latest-released/bin/${arch}" ]; then
+            # Wasn't a version- pick latest
             pathmunge "${CTRLENV_APPS}/${target}/latest-released/bin/${arch}"
             return 0
+        else
+            # Error state: let's try to be a little bit helpful
+            echo "Found matching apps series ${CTRLENV_APPS}/${target} but missing one of arch or version match or no latest-released" >&2
+            return 1
         fi
     fi
     echo "No matching bundle, env, or app found for ctrlenv-pathmunge $1" >&2
@@ -176,10 +184,14 @@ ctrlenv-activate() {
             if [ -d "${CTRLENV_ENVS}/${tg}/bin/${arch}" ]; then
                 pixi shell --manifest-path "${CTRLENV_ENVS}/${tg}/src/${arch}"
                 return $?
-            else
-            # Wasn't a version- pick latest
+            elif [ -d "${CTRLENV_ENVS}/${tg}/latest-released/src/${arch}" ]; then
+                # Wasn't a version- pick latest
                 pixi shell --manifest-path "${CTRLENV_ENVS}/${tg}/latest-released/src/${arch}"
                 return $?
+            else
+                # Error state: let's try to be a little bit helpful
+                echo "Found matching env series ${CTRLENV_ENVS}/${tg} but missing one of arch or version match or no latest-released" >&2
+                return 1
             fi
         fi
     done
--- a/scripts/ctrlenv_setup.sh
+++ b/scripts/ctrlenv_setup.sh
@@ -1,8 +1,9 @@
 #!/bin/bash
 #
-# Source this script to set shared environment variables and gain access to two shell functions:
+# Source this script to set shared environment variables and gain access to three shell functions:
 # 1. ctrlenv-pathmunge: adds the bin from a bundle, environment, or application to your path at a specific version
 # 2. ctrlenv-activate: puts you into a released pixi environment (pixi shell)
+# 3. ctrlenv-versions: shows you which versions exist for an environment or application
 #
 # Defaults are:
 # - Always the latest release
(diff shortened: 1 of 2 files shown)


scripts/ctrlenv_setup.sh
```diff
@@ -0,0 +1,188 @@
...
+    fi
+    # Check envs second: allow omit ctrlenv- prefix
+    for tg in "$target" "ctrlenv-${target}"; do
+        if [ -d "${CTRLENV_ENVS}/${tg}" ]; then
+            if [ -d "${CTRLENV_ENVS}/${tg}/bin/${arch}" ]; then
+                pathmunge "${CTRLENV_ENVS}/${tg}/bin/${arch}"
+                return 0
+            else
+            # Wasn't a version- pick latest
+                pathmunge "${CTRLENV_ENVS}/${tg}/latest-released/bin/${arch}"
```
KaushikMalapati: Should we check if the path with latest-released exists?

ZLLentz: Yeah, let's go back through and check paths to give clearer failure states

scripts/ctrlenv_setup.sh
```diff
@@ -0,0 +1,236 @@
...
+    for tg in "${target}" "ctrlenv-${target}"; do
+        if [ -d "${CTRLENV_ENVS}/${tg}" ]; then
+            if [ -d "${CTRLENV_ENVS}/${tg}/bin/${arch}" ]; then
+                pathmunge "${CTRLENV_ENVS}/${tg}/bin/${arch}"
+                return 0
+            elif [ -d "${CTRLENV_ENVS}/${tg}/latest-released/bin/${arch}" ]; then
+                # Wasn't a version- pick latest
+                pathmunge "${CTRLENV_ENVS}/${tg}/latest-released/bin/${arch}"
+                return 0
+            else
```
KaushikMalapati, line 129: Is it possible that there is an app and environment with the same name, and even though there is a matching app there is no matching env so we return in this case?

ZLLentz: I think this is possible

Maybe we should head this off by enforcing a naming convention for envs and requiring the full name?

e.g. you must call your env `<something>env-<topic>` or even mandate `ctrlenv-`?
And then I remove the automatic translation of e.g. `lucid` -> `ctrlenv-lucid` and avoid a potential conflict

Otherwise I think this gets a bit complicated to handle all the error cases in bash, when I'd like these to be as simple as possible.

KaushikMalapati: A naming convention is reasonable. My first thought is for all envs be in the form <something>env or ctrlenv<something>env so you have to type the env when using ctrlenv-pathmunge but I don't feel strongly about it.

KaushikMalapati (dismissed):
Tried all the commands on psbuild-rhel7, psbuild-rocky9 and xpp-control. Ran into 
```
 WARN the lock file is up-to-date but uses an older format (v6), run `pixi lock` to upgrade to v7 for improved reproducibility
Error:   × failed to collect prefix records from '/reg/g/pcds/pyps/pixi/ctrlenv-external/v0.2.0/src/rhel9-x86_64/.pixi/envs/default'
  ╰─▶ Permission denied (os error 13)
  help: try `pixi clean` to reset the environment and run the command again
```
and couldn't run `pixi lock` to try fixing it because I only have read permissions in the dir (I could have used sudo from an ioc host but I assume there's something in ctrlenv-core that will relax permissions and update the lock file).

Everything else works as expected, having the versions printed out when there's no latest-released seemed pretty helpful.

scripts/ctrlenv_setup.sh
```diff
@@ -0,0 +1,236 @@
...
+        target=base
+    else
+        target="$1"
+    fi
+    arch="$(_ctrlenv-arch)"
+    # For envs allow omit ctrlenv- prefix
+    for tg in "${target}" "ctrlenv-${target}"; do
+        if [ -d "${CTRLENV_ENVS}/${tg}" ]; then
+            if [ -d "${CTRLENV_ENVS}/${tg}/bin/${arch}" ]; then
+                pixi shell --manifest-path "${CTRLENV_ENVS}/${tg}/src/${arch}"
```
ZLLentz: Note to self: does this fail if the user never ran `ctrlenv-pathmunge` and therefore doesn't actually have a pixi executable on the path?

ZLLentz: I think it does fail, and I should add a check for pixi that tells the user to run pathmunge first? Or should I maybe pick a default pixi to use? Pondering

KaushikMalapati: It does fail, although I didn't think there needed to be a special warning when I saw it originally which is why I didn't mention it. I'd opt for telling the user to run pathmunge so that we don't run in to the default pixi not being compatible with whatever you're trying to do.

scripts/ctrlenv_setup.sh
```diff
@@ -0,0 +1,158 @@
...
+    done
+    echo "No matching path or version found for ctrlenv-pathmunge $target" >&2
+    return 1
+}
+
+# Show which versions exist
+ctrlenv-versions() {
+    local target
+    if [ -z "$1" ]; then
+        target=base
```
ZLLentz: oops wrong default

scripts/ctrlenv_setup.sh
```diff
@@ -0,0 +1,158 @@
...
+    done
+    echo "No matching path or version found for ctrlenv-pathmunge $target" >&2
+    return 1
+}
+
+# Show which versions exist
+ctrlenv-versions() {
+    local target
+    if [ -z "$1" ]; then
+        target=ctrlenv-base
```
KaushikMalapati, line 146: Does this exist yet?

KaushikMalapati (approved):
Tried it again, I think it's good

scripts/ctrlenv_setup.sh
ZLLentz: It does not!
I considered removing it and having a different or no default but I think it's OK for this to fail a bit until ctrlenv-base is relased.

KaushikMalapati: Just making sure

PR #336, opened 2026-05-29
ENH: add ctrlenv activation script

ZLLentz:
## Description

Add `ctrlenv-activate` to `ctrlenv_setup.sh` that can be used to activate the new environments in the current shell.

This requires a special activate script to be placed in each environment, which I will do as part of the deployment scripts. The upside of doing it this way is that there aren't any external dependencies for activation, and activation is extremely quick because only environment variables are being set.

Note that, unlike `conda` environments, these cannot be deactivated. You might have some limited success with changing from env to env, but you cannot trivially go back to "before" the activation. If there's some desire for deactivation we can try to re-integrate `pixi shell` for interactive use.

This was previously part of #335 but had some issues:
- Required `pixi` on the path
- Entered a new subshell, preventing its use in scripts
- Didn't work at all

Before merging, I need to:

- [x] Re-run build script on environments released so far to add the activation script
- [x] Release and deploy ctrlenv-base so there's some useful default environment to try this on
- [x] Test thoroughly

I'll ask for reviews after testing.

## Motivation and Context

People want to be able to activate environments sometimes

## How Has This Been Tested?

Interactively only, in all currently deployed environments:
```
ctrlenv-activate ctrlenv-base
ipython
which python
```
```
ctrlenv-activate ctrlenv-widgets
designer
pydm --help
which python
```
```
ctrlenv-activate ctrlenv-lucid
typhos --help
which python
```
```
ctrlenv-activate ctrlenv-external
uv --help
which gh
```

Docstrings, readme

### DIFF
--- a/README.md
+++ b/README.md
@@ -108,14 +108,49 @@ Usage:   /reg/g/pcds/engineering_tools/latest/scripts/check_host HOSTNAME<br/>
 usage:   source ctrlenv_setup.sh<br/>
     <br/>
     Set environment variables and add commands for using ctrlenv python tools.</br>
+    There are three subcommands that become available after sourcing:</br>
+    <ul>
+    <li>ctrlenv-pathmunge: adds a bundle, environment export bin, or application to your path.</li>
+    <li>ctrlenv-activate: activates a python environment.</li>
+    <li>ctrlenv-versions: shows which versions exist for an app or environment.</li>
+    </ul>
     </br>
-    Some examples for subcommands:</br>
-    ctrlenv-pathmunge latest-released</br>
-    ctrlenv-pathmunge 2026/05/26_00</br>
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -114,6 +114,7 @@ usage:   source ctrlenv_setup.sh<br/>
     ctrlenv-pathmunge 2026/05/26_00</br>
     ctrlenv-pathmunge ctrlenv-widgets/v0.1.1</br>
     ctrlenv-pathmunge pytmc</br>
+    ctrlenv-activate ctrlenv-lucid</br>
```
vespos: Should we add the base env example here too?

vespos (dismissed):
LGTM!
Just a very small comment

README.md
ZLLentz: Let's add a few more examples in general

PR #337, opened 2026-05-29
BUG: smartact PV check was always True. Now evaluates correctly if th…

vespos:
This pull request makes a minor update to the `scripts/motor-expert-screen` script, specifically improving the conditional logic used to detect SmarAct motors.

- Logic improvement for motor type detection:
In the previous case, the part `$(caget "$PREFIX:PTYPE_RBV" > /dev/null 2>&1) -eq 0` would always be true. THis now properly checks if the caget exits with no error code

Check with an aerotech and smaract motor, which tests both cases:
```
[xppopr@xpp-control  ~]$ motor-expert-screen XPP:ENS:01:m0
[xppopr@xpp-control  ~]$ motor-expert-screen XPP:LIB:MCS:01
```


ZLLentz (approved):
Shell scripts are super tricky sometimes

PR #338, opened 2026-06-17
Mnt/xpp he

vespos:
XPP is now LCLS-2 for takepeds
Run shellcheck
Update default camera for camViewer

### DIFF
--- a/scripts/camViewer
+++ b/scripts/camViewer
@@ -291,7 +291,7 @@ if [ "$CAMNUM" -eq 0 ]; then
     elif [ "$hutch" = "kfe" ]; then
         CAMNAME=IM1K0
     elif [ "$hutch" = "xpp" ]; then
-        CAMNAME=hx2_sb1_yag
+        CAMNAME=IM1L2-01
     elif [ "$hutch" = "xcs" ]; then
         CAMNAME=xcs_yag5
     elif [ "$hutch" = "mfx" ]; then
--- a/scripts/takepeds
+++ b/scripts/takepeds
@@ -10,12 +10,12 @@ Takes a run with dark images for use in pedestals, and posts to the elog.
 EOF
 }
 
-if [[($1 == "--help") || ($1 == "-h")]]; then
+if [[ ($1 == "--help") || ($1 == "-h") ]]; then
         usage
         exit 0
 fi
 
-EXP=`get_curr_exp`
+EXP=$(get_curr_exp)
 HUTCH=${EXP:0:3}
 
 #SIT_ENV_DIR="/sdf/group/lcls/ds/ana/sw"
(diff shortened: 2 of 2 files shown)


scripts/camViewer
```diff
@@ -285,7 +285,7 @@ if [ "$CAMNUM" -eq 0 ]; then
     elif [ "$hutch" = "kfe" ]; then
         CAMNAME=IM1K0
     elif [ "$hutch" = "xpp" ]; then
-        CAMNAME=hx2_sb1_yag
+        CAMNAME=IM1L2-01
```
ZLLentz, line 288: Undocumented change but it makes sense

scripts/takepeds
```diff
@@ -25,37 +25,40 @@ HUTCH=$(get_info --gethutch)
...
-        epixquad_pedestal_scan --record 1 --detname epixquad1kfps_0 --hutch ${HUTCH}
+        epixquad_pedestal_scan --record 1 --detname epixquad1kfps_0 --hutch "${HUTCH}"
     elif [[ ${HUTCH} =~ 'mfx' ]]; then
-        jungfrau_pedestal_scan --record 1 --hutch ${HUTCH} -p 0 -g 1 -v -t 10000 -C drp-srcf-cmp014
+        jungfrau_pedestal_scan --record 1 --hutch "${HUTCH}" -p 0 -g 1 -v -t 10000 -C drp-srcf-cmp014
     elif [[ ${HUTCH} =~ 'rix' ]]; then
         echo "Running LCLS2 RIX specific pedestal acquisition..."
         timed_run --duration="60" --config="/cds/group/pcds/dist/pds/rix/scripts/rix.py" --record
+    elif [[ ${HUTCH} =~ 'xpp' ]]; then
+        jungfrau_pedestal_scan --record 1 --hutch "${HUTCH}" --detname jungfrau1M -C drp-srcf-mon008 --nprocs 1
```
ZLLentz, line 37: I wonder at what point we need to organize/track this behavior differently, or if having a separate if clause for each instrument tracked here will always make sense

ZLLentz (approved):
I trust you tested this
My only request would be to update the PR description to include the default cam change just to document it

scripts/takepeds
silkenelson: https://jira.slac.stanford.edu/browse/ECS-6777
states that we have asked to get the LCLS1 behavior back. Hardcoding a single detector/hutch does already not work well when we deploy a mix of detectors (epixM & AndOR for RIXS,...) and it's on the to-do list for the data team.

ZLLentz: I'm thinking more along the lines of: if there isn't a better way, can we hardcode a path to e.g. a script that can be changed by the cdso without needing to make an engineering tools PR?

vespos: We will need to revise this soon because of the DAQ update anyway... Let's follow up then

PR #340, opened 2026-07-07
ENH: add ctrlenv-shell and document it

ZLLentz:
## Description

Add one more ctrlenv helper function: `ctrlenv-shell`

This activates a centrally installed pixi environment in a exitable/deactivatable subshell
Unlike `ctrlenv-activate`, which cannot be deactivated, this can be deactivated.
Unlike `ctrlenv-activate`, which is safe to include in a script, this cannot be included in a script.

The behavior is extremely similar to the `pixi shell` function, but implementing it this way means the user isn't required to have pixi on their path to use `ctrlenv-shell`.

I've also fixed some typos in the script docstring.

## Motivation and Context

@aberges-SLAC  requested a tool that would allow him to deactivate the pixi environments, as we used to be able to do with conda. This should satisfy those requirements.

## How Has This Been Tested?

Interactively only.

In comments and the README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -108,10 +108,11 @@ Usage:   /reg/g/pcds/engineering_tools/latest/scripts/check_host HOSTNAME<br/>
 usage:   source ctrlenv_setup.sh<br/>
     <br/>
     Set environment variables and add commands for using ctrlenv python tools.</br>
-    There are three subcommands that become available after sourcing:</br>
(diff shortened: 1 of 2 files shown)


KaushikMalapati (approved):

yannS2016 (approved):
Great feature here, answering some of my earlier comments.

PR #341, opened 2026-07-20
adding logic for tc motor

ctslac:
## Description

Added support for beckhoff motors with the RTYP as tcmotor.
## Motivation and Context

The user motors pvs have this value
## How Has This Been Tested?

I tested on my checkout.

### DIFF
--- a/scripts/motor-expert-screen
+++ b/scripts/motor-expert-screen
@@ -39,11 +39,11 @@ if [ "${rtyp}" == 'xps8p' ]; then
     title='Newport XPS Positioner'
 elif [ "${rtyp}" == 'ims' ]; then
     title='IMS Motor Control -- Main'
-elif [ "${rtyp}" == 'motor' ]; then
+elif [[ "${rtyp}" == 'motor' ]] || [[ "${rtyp}" == 'tcmotor' ]]; then
     # Check to see if this is a Beckhoff axis
     #   caget will return exit code 0 if :PLC:nErrorId_RBV exists,
     #   we set isbeckhoff based on this.
-    if caget "${PREFIX}:PLC:nErrorId_RBV" > /dev/null 2>&1; then
+    if [[ "${rtyp}" == 'tcmotor' ]] || caget "${PREFIX}:PLC:nErrorId_RBV" > /dev/null 2>&1; then
         title="Typhos Suite - ${PREFIX}"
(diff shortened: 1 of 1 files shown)


scripts/motor-expert-screen
```diff
@@ -39,11 +39,11 @@ if [ "${rtyp}" == 'xps8p' ]; then
     title='Newport XPS Positioner'
 elif [ "${rtyp}" == 'ims' ]; then
     title='IMS Motor Control -- Main'
-elif [ "${rtyp}" == 'motor' ]; then
+elif [[ "${rtyp}" == 'motor' ]] || [[ "${rtyp}" == 'tcmotor' ]]; then
     # Check to see if this is a Beckhoff axis
     #   caget will return exit code 0 if :PLC:nErrorId_RBV exists,
     #   we set isbeckhoff based on this.
-    if caget "${PREFIX}:PLC:nErrorId_RBV" > /dev/null 2>&1; then
+    if [[ "${rtyp}" == 'tcmotor' ]] || caget "${PREFIX}:PLC:nErrorId_RBV" > /dev/null 2>&1; then
```
ZLLentz, line 46: Did you make sure this still works for non tcmotor beckhoff motors?

ctslac: Yeah, regular IMS motors still work; I haven't tested on Aerotech yet. I'm not sure if we still use them.

ZLLentz: I don't mean to ask you to test every motor type, just the ones affected by this line: motors on beckhoff IOCs that aren't `tcmotor` type

ctslac: Yeah, I wanted to check Aerotech becuase I see the beckhoff logic is close the Aerotech. But it still works

ZLLentz: Ah yeah I see now that aerotech is the `else` clause here- didn't see it on review because it was outside the diff. Good job being thorough.

ZLLentz: But yeah I'll approve this if you can confirm that "old" beckhoff motors still work

ctslac: Yep! Tc motors with the RTYP as 'motor' still work.

ZLLentz (approved):

PR #348, opened 2026-08-21
Adds a script for extracting the eerom serial number for SLAC evrs/tprs

ddamiani:
## Description
Adds a script for extracting the eerom serial number for SLAC evrs/tprs. If you run this script it will print out the serial numbers of all the gen1 and gen2 evr cards in the machine. It works with either the evr or tpr firmware. It lists them in the order they are enumerated by the pcie bus this should map to how the kernel module enumerates them. e.g. the first is era/tpra, second is erb/tprb, etc.

## Motivation and Context
So we can inventory and keep track of what physical cards went where.

```
ddamiani@daq-cxi-master:~$ sudo evrserialnum
78-1d-84-70-f1-08-5c-00
28-1d-84-70-f1-08-5c-00
62-1d-84-70-f1-08-5c-00
```


silkenelson (approved):
This looks great and I've seen the test 'snapshots'

